PR #12, opened 2017-11-17
Greg/swagger: Refactor DispatchJob V1 implementation

silverspace:
Refactoring the previous DispatchJobs implementation to reduce copy/paste and set up a template for defining V2 DispatchJobs. 

This closely models the structs defined in fleet_dispatch.go:
 * DispatchJobV1Base => DispatchJobV1Create (with the exception of "id")
 * DispatchJobV1Response => DispatchJobV1 (extends DispatchJobV1Create and adds "id", "group_id", "org_id", and "job_state")

Tested locally by using Charles to rewrite the URL in the body of samsara.com/api to point to my raw github file. Verified that the page still loads correctly.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1046,7 +1046,7 @@
                     "200": {
                         "description": "Dispatch jobs for the specified group.",
                         "schema": {
-                            "$ref": "#/definitions/DispatchJobsResponse"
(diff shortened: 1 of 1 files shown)


silverspace:
Thanks!

PR #13, opened 2017-11-21
swagger: Add support for dispatch routes

silverspace:
This PR adds support for all of the dispatch route REST endpoints in our swagger.json file. 

The routes added came from samsaradev.io/controllers/api/routes.go, with the exception of 
  * /fleet/drivers/{driver_id}/dispatch/routes/{route_id}
  * /fleet/vehicles/{vehicle_id}/dispatch/routes/{route_id}

For now, those will be unexported functions that only we know about, since exposing them may add unnecessary complexity.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2356,6 +2356,132 @@
                 }
             }
         },
+        "DispatchJobCreate": {
+            "type": "object",
+            "required": [
+                "scheduled_arrival_time_ms",
+                "destination_lat",
+                "destination_lng"
+            ],
+            "properties": {
+                "driver_id": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description":  "ID of the driver assigned to the dispatch job.",
+                    "example": 444
+                },
+                "vehicle_id": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "ID of the vehicle used for the dispatch job.",
+                    "example": 112
+                },
+                "scheduled_arrival_time_ms": {
+                    "type": "integer",
+                    "format": "int64",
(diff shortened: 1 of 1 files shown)


silverspace:
We are merging this now so that we will have the public documentation updated for a meeting with Bolt today. Please do add comments to this PR or the commits and I will address them in a follow-up PR.

swagger.json
```diff
@@ -2494,6 +3082,70 @@
...
+            "schema": {
+                "$ref": "#/definitions/DispatchRouteCreate"
+            }
+        },
+        "routeUpdateParam": {
+            "name": "updateDispatchRouteParams",
+            "required": true,
+            "in": "body",
+            "schema": {
+                "$ref": "#/definitions/DispatchRoute"
```
YosubShin: I think we should have a separate schema definition for `DispatchRouteUpdate` since it doesn't support setting `job_state`, `arrived_at_ms`, `completed_at_ms`, `skipped_at_ms`.

silverspace: When we ask consumers to PUT a route, we should expect them to send us exactly what we sent them in a GET request, with some values modified. I think we should still expect them to populate completed_at_ms even if they can't modify the value.

YosubShin: I think Go JSON marshaler would ignore additional fields, so it's ok for users to send us extra fields noted above. 
However I think it would be confusing to the users of the API if we let the users specifying fields like `job_state` but they wouldn't have any effect.
Also, if someone manually construct an API request, I personally think it's nicer to not requiring them to specify extra fields that are functionally not necessary.

PR #16, opened 2017-12-11
swagger: Remove driver_id and vehicle_id from JobCreateParam

silverspace:
Moving driver_id and vehicle_id from the Job creation parameters to the
Route creation parameters, since we don't allow specifying driver_id and
vehicle_id on a per-job basis.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2386,18 +2386,6 @@
                 "destination_lng"
             ],
             "properties": {
-                "driver_id": {
-                    "type": "integer",
-                    "format": "int64",
-                    "description":  "ID of the driver assigned to the dispatch job.",
-                    "example": 444
-                },
-                "vehicle_id": {
-                    "type": "integer",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -2386,18 +2386,6 @@
                 "destination_lng"
             ],
             "properties": {
-                "driver_id": {
```
YosubShin: Can we also update route creation/update endpoints saying only one of `driver_id` or `vehicle_id` should be populated?
Right now, the wording is slightly misleading and could be interpreted as it's ok to populate both `driver_id` and `vehicle_id`.

>  vehicle_id: ID of the vehicle assigned to the dispatch route. One of driver_id OR vehicle_id is required.
> driver_id: ID of the driver assigned to the dispatch route. One of driver_id OR vehicle_id is required.

silverspace: Sure, updated.

PR #18, opened 2018-02-21
swagger: Add documentation for asset-specific endpoints

arthurhuang:
This adds documentation for:
/v1/fleet/assets
/v1/fleet/assets/{assetId:[0-9]+}/reefer

and introduces a new "Assets" tag

Verified on http://editor.swagger.io/#

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -36,6 +36,10 @@
         {
             "name": "Sensors",
             "description": "Access to sensor data"
+        },
+        {
+            "name": "Assets",
+            "description": "Access to asset-specific data"
         }
     ],
     "paths": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -3119,6 +3196,215 @@
...
+                "cable": {
+                    "type": "array",
+                    "description": "The cable connected to the asset",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "assetType": {
+                                "type": "string",
+                                "description": "Asset type",
+                                "example": "Reefer (Thermo King)"
```
kl4kennylee81: should we be giving more examples? I think it might make sense for us internally but may be more difficult for a user to understand what are the different types of assets

swagger.json
```diff
@@ -3119,6 +3196,215 @@
...
+                        "alarms": {
+                            "type": "array",
+                            "description": "Reefer alarms",
+                            "items": {
+                                "type": "object",
+                                "properties": {
+                                    "changedAtMs": {
+                                        "type": "integer",
+                                        "format": "int64",
+                                        "description": "Timestamp in Unix milliseconds since epoch.",
```
kl4kennylee81: "The time in Unix epoch milliseconds when the alarms were last polled." hmm not sure what would be a more thorough explanation

swagger.json
```diff
@@ -3119,6 +3196,215 @@
...
+                                        "example": 1453449599999
+                                    },
+                                    "setPointInMilliC": {
```
kl4kennylee81: just a small observation should be change this to temperatureInMilliC. nbd either would be fine.

swagger.json
arthurhuang: i don't think the docs allow for multiple examples, but if there's confusion from the users we can look into explaining the options further in the description

swagger.json
arthurhuang: i'll change it to `Timestamp in Unix milliseconds since epoch when the alarms were reported.`

swagger.json
arthurhuang: ok

PR #19, opened 2018-02-22
swagger.json: add in asset location api

kl4kennylee81:
This change adds the asset location api to our public api.

The routes added came from samsaradev.io/controllers/api/assets.go

/fleet/asset/{assetId: [0-9]+}/location

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1742,14 +1742,7 @@
                 "operationId": "GetAssetReefer",
                 "parameters": [
                     { "$ref": "#/parameters/accessTokenParam" },
-                    {
-                        "name": "asset_id",
-                        "in": "path",
-                        "required": true,
-                        "description": "ID of the asset",
-                        "type": "integer",
-                        "format": "int64"
-                    },
+                    { "$ref": "#/parameters/assetIdParam" },
                     { "$ref": "#/parameters/routeHistoryStartTimeParam" },
                     { "$ref": "#/parameters/routeHistoryEndTimeParam" }
                 ],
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -1768,6 +1768,43 @@
...
+                "tags": [
+                    "Default", "Fleet", "Assets"
+                ],
+                "summary": "/fleet/assets/{assetId:[0-9]+}/location",
+                "description": "Fetch the historical locations for the asset.",
+                "operationId": "GetAssetLocation",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "asset_id",
```
arthurhuang: this should probably be a `$ref` in `#/parameters/` since we'll use it a bunch

swagger.json
```diff
@@ -3405,6 +3442,37 @@
...
                 }
             }
+        },
+        "AssetLocationResponse": {
+            "type": "object",
+            "description": "Asset location details.",
+            "properties": {
+                "location": {
+                    "type": "string",
+                    "description": "The best matched (street,city,state) for the latititude and longitude.",
```
arthurhuang: "best effort" instead of "best matched"

swagger.json
```diff
@@ -3405,6 +3442,37 @@
...
+                    "example": "37"
+                },
+                "longitude": {
+                    "type": "number",
+                    "description": "The longitude of the location in degrees.",
+                    "example": "-122"
+                },
+                "speedMilesPerHour": {
+                    "type": "number",
+                    "description": "The speed calculated from the GPS that the asset was traveling at in miles per hour.",
```
arthurhuang: `from GPS`

swagger.json
kl4kennylee81: is this basically making a variable parameter? i'll try my best swagger skills but don't have high hopes

kl4kennylee81: done

swagger.json
kl4kennylee81: done

swagger.json
kl4kennylee81: done

swagger.json
```diff
@@ -1768,6 +1761,36 @@
                     }
                 }
             }
+        },
+        "/fleet/assets/{asset_id}/location": {
```
arthurhuang: This should be `locations` plural, because we're returning the list of locations for the time period. Should change the actual API & test too

swagger.json
```diff
@@ -3405,6 +3428,37 @@
                     }
                 }
             }
+        },
+        "AssetLocationResponse": {
+            "type": "object",
```
arthurhuang: The response is an array of these objects, instead of just 1 object

swagger.json
```diff
@@ -3549,6 +3603,14 @@
             "in": "query",
             "type": "integer",
             "format": "int64"
+        },
+        "assetIdParam": {
```
arthurhuang: thanks for changing this

swagger.json
kl4kennylee81: done

swagger.json
kl4kennylee81: lol good catch

PR #30, opened 2018-06-12
Jrsaunde/industrial api

jrsaunde:
This adds documentation for two new api endpoints:

/industrial/data => list all data inputs in a group (and optional historical datapoints):
screen shot 2018-06-11 at 7 36 04 pm

/industrial/data/{data_input_id} =>  fetch datapoints for a dataInput
screen shot 2018-06-11 at 7 36 25 pm

This also moves the machine endpoints under the industrial tag, put doesn't change the path, so current users can still use the API:

screen shot 2018-06-11 at 7 37 06 pm

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1804,6 +1804,42 @@
                 }
             }
         },
+        "/industrial/data/{data_input_id}": {
+            "get": {
+                "tags": [
+                    "Industrial", "Default"
+                ],
+                "summary": "/industrial/data/{data_input_id:[0-9]+}",
+                "description": "Fetch datapoints from a given data input.",
+                "operationId": "GetDataInput",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/dataInputIdParam"},
+                    { "$ref": "#/parameters/dataInputStartTimeParam"},
+                    { "$ref": "#/parameters/dataInputEndTimeParam"}
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Returns datapoints for the given data input",
+                        "schema": {
+                            "$ref": "#/definitions/DataInputHistoryResponse"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
         "/tags": {
             "get": {
                 "tags": [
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4350,6 +4432,21 @@
             "description": "ID of the asset",
             "type": "integer",
             "format": "int64"
+        "dataInputEndTimeParam": {
+            "name": "endMs",
+            "description": "Timestamp in milliseconds representing the end of the period to fetch, inclusive. Used in combination with startMs.",
```
silverspace: Nit: "timestamp in *Unix* milliseconds"

swagger.json
```diff
@@ -4350,6 +4432,21 @@
...
+            "required": false,
+            "in": "query",
+            "type": "integer",
+            "format": "int64"
+        },
+        "dataInputStartTimeParam": {
+            "name": "startMs",
+            "description": "Timestamp in milliseconds representing the start of the period to fetch, inclusive. Used in combination with endMs.",
```
silverspace: We should also call out the default value.

swagger.json
```diff
@@ -3760,6 +3804,44 @@
...
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "timeMs":{
+                                "type":"integer",
+                                "format": "int64",
+                                "example": 1453449599999
+                            },
+                            "value":{
+                                "type":"integer",
```
silverspace: Should type be "number" instead of "integer"? The integer + float combination looks a little suspect.

swagger.json
```diff
@@ -1804,6 +1804,42 @@
                 }
             }
         },
+        "/industrial/data/{data_input_id}": {
+            "get": {
+                "tags": [
+                    "Industrial", "Default"
+                ],
+                "summary": "/industrial/data/{data_input_id}",
```
silverspace: In other API endpoints we define bind variables as `{data_input_id:[0-9]+}`

swagger.json
```diff
@@ -1804,6 +1804,42 @@
...
+                ],
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/dataInputIdParam"},
+                    { "$ref": "#/parameters/dataInputStartTimeParam"},
+                    { "$ref": "#/parameters/dataInputEndTimeParam"}
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Data points from this data input",
```
silverspace: The `/industrial/data` endpoint uses the noun `data inputs`. We should be consistent with this endpoint.

swagger.json
```diff
@@ -4432,6 +4468,15 @@
             "description": "ID of the asset",
             "type": "integer",
             "format": "int64"
+        },
+        "dataInputIdParam": {
+            "name": "data_input_id",
+            "in": "path",
+            "required": true,
+            "description": "ID of the data_input",
```
silverspace: Nit: data_input is not a variable anywhere, we can just call this "data input".

swagger.json
jrsaunde: done!

swagger.json
jrsaunde: done!

swagger.json
jrsaunde: done!

swagger.json
jrsaunde: done!

PR #31, opened 2018-06-14
swagger:Updated description on Put /tags/{tag_id}; Updated descriptio…

mnn-agarwal:
…n on /fleet/drivers/create to mention that password is not a required field. Have confirmed the file has no issues on editor.swagger.io


swagger.json
```diff
@@ -1351,6 +1347,41 @@
                 }
             }
         },
+        "/fleet/dispatch/routes/job_updates": {
+            "get": {
+                "tags": [
+                    "Fleet", "Routes", "Default"
```
silverspace: Could we make Routes the first in the list?

PR #34, opened 2018-07-09
swagger:Updated hos_logs response to include remarks in the return body

mnn-agarwal:
Verified on editor.swagger.io

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -3267,8 +3267,8 @@
                             },
                             "remark":{
                                 "type": "string",
-                                "description": "Remarks associated with the log entry.",
-                                "example": "Foo Bar"
+                                "description": "Remark associated with the log entry.",
+                                "example": "Lunch Break"
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -3264,6 +3264,11 @@
                                 "type": "string",
                                 "description": "Name of location at which the log was recorded.",
                                 "example": "McLean Site A"
+                            },
+                            "remark":{
+                                "type": "string",
+                                "description": "Remarks associated with the log entry.",
+                                "example": "Foo Bar"
```
silverspace: Let's use a more realistic example here, such as "Lunch break".

swagger.json
```diff
@@ -3264,6 +3264,11 @@
                                 "type": "string",
                                 "description": "Name of location at which the log was recorded.",
                                 "example": "McLean Site A"
+                            },
+                            "remark":{
+                                "type": "string",
+                                "description": "Remarks associated with the log entry.",
```
silverspace: Could we use "Remark" instead of "Remarks", to keep the plurality consistent with the key name?

PR #36, opened 2018-07-09
swagger:Modified description to be more descriptive and provide detai…

mnn-agarwal:
…ls around how to obtain API keys, etc.

Checked on editor.swagger.io


swagger.json
```diff
@@ -3,7 +3,7 @@
...
...he error, you can refer to the following table for understanding the error message.\n\n| Status Code | Message | Description |\n|-------------|----------------|-------------------------------------------------------------------|\n| 401 | Invalid token | The API token is invalid and could not be authenticated. Please refer to the [authentication section](#section/Authentication). |\n| 404 | Page not found | The API endpoint being accessed is invalid. |\n| 400 | Bad request | Default response for an invalid request. Please check the request to make sure it follows the format specified in the documentation. |\n\n# Versioning\n\nAll our APIs are versioned. Our current API version is `v1` and we are continuously working on improving it further and provide additional endpoints. If we intend to make breaking changes to an API which either changes the response format or request parameter, we will increment the version. Thus, you can use our current API version worry free.\n\n# FAQs\n\nCheck out our [responses to FAQs here](https://kb.samsara.com/hc/en-us/sections/360000538054-APIs). Don’t see an answer to your question? Reach out to us on [support@samsara.com](mailto:support@samsara.com)."
```
silverspace: Hmm, this introduction section isn't very maintainable if someone wants to make an update. Does the rendered swagger file look correct if we add newlines into the JSON string?

mnn-agarwal: It does not :(
If we add newline in the json it breaks the file. We might want to move to YAML at some point. that is more readable

silverspace: I tested this on editor.swagger.com and it ignores any newlines that we add. For example:

```
"description": "# Introduction\n\n
Samsara provides API endpoints...
```

silverspace: Nevermind, JSON does not allow \A newline characters. This is a bummer, but I guess we won't be updating this too often.

PR #37, opened 2018-07-10
swagger: add activeMs

dsharrison:
This is the documentation update for https://github.com/samsara-dev/backend/pull/10589 and will be merged after that change lands.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -3600,6 +3600,12 @@
                     "description": "Hours spent on duty or driving, rounded to two decimal places.",
                     "example": 5.40
                   },
+                  "activeMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "Milliseconds spent on duty or driving.",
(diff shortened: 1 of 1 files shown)


btuan:
Looks reasonable to me. Just for sanity-check, can you run the changed file through a swagger linter?
Maybe something like this?

https://editor.swagger.io/

silverspace:
This looks fine to me. A higher level question: Do we want to expose sub-minute level precision to external customers for HOS data?

silverspace:
I'm not sure about the data inaccuracies, that's certainly a possibility. I think the HOS app displays data with a granularity of a minute, so my question was whether we want to mirror the same precision in our APIs. 

In any case, I'm fairly certain we have precedent of using millisecond timestamps in our HOS APIs, so this was more of a follow-up question for Compliance about HOS API design.

silverspace:
Should we merge this PR?

PR #47, opened 2018-08-30
swagger: update api docs to include safety score endpoints

ChenJesse:
- adds endpoints for the vehicle safety score API endpoint, and then driver safety score API endpoint
- the `DriverSafetyScoreResponse` and `VehicleSafetyScoreResponse` share all the fields except the driver one will have `driverId` and vehicle will have `vehicleId`, wasn't sure if there was a way to create some structure for all the shared fields and then somehow spread them

Used https://editor.swagger.io/ to validate + test it
safetyscoreapi

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -34,6 +34,10 @@
             "description": "Access to asset specific data"
         },
         {
+            "name": "Safety",
+            "description": "Access to safety specific data"
+        },
+        {
             "name": "Default",
             "description": "All Samsara API endpoints"
         }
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4822,6 +4904,156 @@
...
+                    "type": "integer",
+                    "description": "Harsh acceleration event count",
+                    "example": 1
+                },
+                "harshTurningCount": {
+                    "type": "integer",
+                    "description": "Harsh turning event count",
+                    "example": 0
+                },
+                "crashCount": {
```
silverspace: Does it make sense to share this model definition with `DriverSafetyScoreResponse`?

swagger.json
```diff
@@ -5153,6 +5385,22 @@
             "in": "query",
             "type": "integer",
             "format": "int64"
+        },
+        "safetyScoreStartMsParam": {
+            "name": "startMs",
+            "description": "Timestamp in milliseconds representing the start of the period to fetch, inclusive. Used in combination with endMs.",
```
silverspace: Is there a maximum duration that this endpoint enforces?

swagger.json
ChenJesse: so they are the exact same, except the `VehicleSafetyScoreResponse` has `vehicleId` instead of `driverId`. I wasn't sure if there was a way to consolidate all the other shared properties though...

swagger.json
ChenJesse: there isn't right now, we aren't too worried about expensive queries since we are now precomputing safety score. I will add it in later if we do enforce a max though

PR #48, opened 2018-09-07
api-docs swagger: add new start and end Address field to trips 

AChelikani:
Modify old `startLocation` and `endLocation` fields to mean geocoded address and add new `startAddress` and `endAddress` to replace.


silverspace:
Should we still merge this PR?

AChelikani:
Yeah, I think it still applies.

PR #49, opened 2018-09-07
swagger: add vehicleId to driver docs; correct snake to camel case

alex-wuu:
Added `vehicleId` to driver documents

Incorrectly showed driver doc API response as snake case in swagger, and corrected it here to show the response in camel case instead

screen shot 2018-09-07 at 12 45 27 pm

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -5078,30 +5078,30 @@
                 "$ref": "#/definitions/DocumentField"
             }
         },
-        "Document": {            
+        "Document": {
             "type": "object",
             "required": [
-                "driver_id",
-                "driver_created_at_ms",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -5112,6 +5112,12 @@
                     "description": "ID of the Samsara dispatch job.",
                     "example": 773
                 },
+                "vehicleId": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "Vehicle ID of the driver at document creation.",
```
jenniferjzhang: nit: maybe "VehicleID of the vehicle assigned to the driver at document creation." ?

PR #54, opened 2018-10-03
swagger: add destinationAddressId to DispatchJobs

tirodactyl:
Tested in Swagger editor:
20181002_dispatchjob-destinationaddressid

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -4280,6 +4280,12 @@
                             "format": "int64",
                             "description": "The time at which the assigned driver is estimated to arrive at the job destination. Only valid for en-route jobs.",
                             "example": 1462881998034
+                        },
+                        "destination_address_id": {
+                            "type": "integer",
+                            "format": "int64",
+                            "description": "ID of the job destination associated with an address book entry.",
+                            "example": 67890
                         }
                     }
                 },


swagger.json
```diff
@@ -4163,6 +4163,12 @@
                     "description": "The address of the job destination, as it would be recognized if provided to maps.google.com",
                     "example": "123 Main St, Philadelphia, PA 19106"
                 },
+                "destination_address_id": {
```
kelseylam: Ah, I think you've added this to `DispatchJobCreate`. We do eventually want this to be here, but currently since we don't have a POST endpoint done, maybe we should wait to merge this after that's implemented. 

Also small nit, to be consistent with the descriptions for destination name and destination address, maybe say "The ID of the job destination associated with an address book entry."

swagger.json
```diff
@@ -4280,6 +4280,12 @@
                             "format": "int64",
                             "description": "The time at which the assigned driver is estimated to arrive at the job destination. Only valid for en-route jobs.",
                             "example": 1462881998034
+                        },
+                        "destination_address_id": {
```
alex-wuu: ah we do snake case here lol

swagger.json
tirodactyl: Moved per our conversation.

PR #55, opened 2018-10-04
Update DocumentField with stringValue, numberValue, photoValue

kelseylam:
We updated DocumentField to use `stringValue`, `numberValue`, and `photoValue` as a replacement for the generic type `value`. Only one of these values should be non-nil.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -5099,8 +5099,9 @@
             "type": "object",
             "required": [
                 "label",
-                "value"
+                "valueType"
             ],
+            "description": "A field of a document. A field can be a string type, number type, or photo type, which you can identify from its valueType. Between stringValue, numberValue, and photoValue, only one can exist for a single document field depending on the valueType.",
             "properties": {
                 "label": {
                     "type": "string",
@@ -5108,8 +5109,39 @@
                     "example": "Fuel Cost ($)"
                 },
                 "value": {
-                    "description":  "Value of this field. Depending on what kind of field it is, this may be one of the following: an array of image urls, a float, an integer, or a string.",
-                    "example": 23.45
+                    "description":  "DEPRECATED: Please use stringValue, numberValue, or photoValue instead. Value of this field. Depending on what kind of field it is, this may be one of the following: an array of image urls, a float, an integer, or a string.",
+                    "example": 23.45,
+                    "deprecated": true
+                },
+                "valueType": {
+                    "type": "string",
+                    "description": "Determines the type of this field and what type of value this field has. It should be either ValueType_Number, ValueType_String, or ValueType_Photo.",
+                    "example": "ValueType_Number"
+                },
+                "stringValue": {
+                    "type": "string",
+                    "description": "Value of this field if this document field has valueType: ValueType_String.",
+                    "example": "This is a string."
+                },
+                "numberValue": {
+                    "type": "number",
+                    "format": "double",
+                    "description": "Value of this field if this document field has valueType: ValueType_Number.",
+                     "example": 12.34
+                },
+                "photoValue": {
+                    "type": "array",
+                    "description": "Value of this field if this document field has valueType: ValueType_Photo. Array of photo objects where each object contains a URL for a photo.",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "url": {
+                                "type": "string",
+                                "description": "Photo URL for a JPG image"
+                            }
+                        }
+                    },
+                  "example": [{ "url": "https://www.samsara.com/photo1" }, { "url": "https://www.samsara.com/photo2"}]
                 }
             }
         },


swagger.json
```diff
@@ -5101,15 +5101,41 @@
...
+                "stringValue": {
+                  "type": "string",
+                  "description": "Value of this field if this document field is a string type field.",
+                   "example": "This is a string."
+                },
+                "numberValue": {
+                  "type": "number",
+                  "format": "float",
+                  "description": "Value of this field if this document field is a number type field.",
+                   "example": 12.34
```
sn0v: ooc: do we plan on providing POST endpoints for documents in the future? If so, will the `Document` definition be required to somehow support validation? (eg - number of decimal places)

The current implementation seems good for GET endpoints, but this might be something to think about so we don't end up having to duplicate definitions or deprecate `numberValue` prematurely.

swagger.json
```diff
@@ -5101,15 +5101,41 @@
...
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "url": {
+                        "type": "string",
+                        "description": "Photo URL"
+                      }
+                    }
+                  },
+                  "example": [{ "url": "www.samsara.com/photo1" }, { "url": "www.samsara.com/photo2"}]
```
alex-wuu: nit: add `https://` to the url examples here?

swagger.json
```diff
@@ -5101,15 +5101,41 @@
...
+            "description": "A field of a document. A field can be a string type, number type, or photo type. Between stringValue, numberValue, and photoValue, only one can exist for a single document field depending on the field's type.",
             "properties": {
                 "label": {
                     "type": "string",
                     "description": "Descriptive name of this field.",
                     "example": "Fuel Cost ($)"
                 },
                 "value": {
-                    "description":  "Value of this field. Depending on what kind of field it is, this may be one of the following: an array of image urls, a float, an integer, or a string.",
+                    "description":  "DEPRECATED: Please use stringValue, numberValue, or photoValue instead. Value of this field. Depending on what kind of field it is, this may be one of the following: an array of image urls, a float, an integer, or a string.",
```
alex-wuu: I couldn't find any other examples where we do this, but do we have a standardized way to deprecate API fields?

silverspace: It's fine to add `DEPRECATED` to the description, but please also add the `"deprecated": true` OpenAPI 3.0 annotation to this value.

swagger.json
```diff
@@ -5101,15 +5101,41 @@
...
                     "example": 23.45
+                },
+                "stringValue": {
+                  "type": "string",
+                  "description": "Value of this field if this document field is a string type field.",
+                   "example": "This is a string."
```
silverspace: Nit: whitespace formatting is off in GitHub.

swagger.json
```diff
@@ -5101,15 +5101,41 @@
...
+                },
+                "numberValue": {
+                  "type": "number",
+                  "format": "float",
```
silverspace: I suspect this should be double instead of float.

swagger.json
```diff
@@ -5101,15 +5101,41 @@
...
+                },
+                "photoValue": {
+                  "type": "array",
+                 "description": "Value of this field if this document field is a photo type field. Array of photo objects where each object contains a URL for a photo.",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "url": {
+                        "type": "string",
+                        "description": "Photo URL"
```
silverspace: What type of photo is this expected to return? (JPG, PNG, etc).

silverspace: Does a photo not have a name or description associated with it?

swagger.json
kelseylam: Ooh, great I didn't know this existed!

swagger.json
kelseylam: Yes. Once we hit the graphql endpoint, that will handle validations for num decimal places. Also, presumably they should fetch the document template to see how many decimal places they should use for this value.

sn0v: Nvm, talked in person. Seems like we don't plan on re-using this def for document template POST, so the lack of validation shouldn't be an issue.

swagger.json
sn0v: I think adding the type of the returned photo (jpg) to the description would be worthwhile. Are there any downsides to exposing the UUID?

silverspace: I hope not, since I assume the URL will include the `uuid` in the path. Is there also a human assigned name / description to each photo? In other words, can the driver annotate each photo?

kelseylam: I don't think there are any downsides. There is not a human assigned name for the photo.

swagger.json
kelseylam: When I add `"deprecated": true` under `"value"`, it says "should not have additional properties additionalProperties: deprecated" on editor.swagger.io. Do you think the editor is just wrong?

PR #59, opened 2018-10-09
Add destination_address_id to DispatchJobCreate def

sn0v:
Adds `destination_address_id` to our swagger docs for POST/PUT routes.

This should be merged after https://github.com/samsara-dev/backend/pull/14880 deploys.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -4194,6 +4194,12 @@
                     "description": "The address of the job destination, as it would be recognized if provided to maps.google.com",
                     "example": "123 Main St, Philadelphia, PA 19106"
                 },
+                "destination_address_id": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "ID of the job destination associated with an address book entry.",
+                    "example": 67890
+                },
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4194,6 +4194,12 @@
                     "description": "The address of the job destination, as it would be recognized if provided to maps.google.com",
                     "example": "123 Main St, Philadelphia, PA 19106"
                 },
+                "destination_address_id": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "ID of the job destination associated with an address book entry.",
```
alex-wuu: nit: add "The" at the beginning of this description for consistency? We have things like `"description": "The ID of this tag."` in other places

sn0v: Good point. I'm inclined to keep it as is because I'm seeing it done both ways in our docs (lol), but also because most of the dispatch/driver docs api docs start with `ID of ...`

alex-wuu: seems fine to me then

PR #62, opened 2018-10-12
swagger.json: add api docs for addresses CRUD

ehzhang:
Here we go!

image
image
image
image
image

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2574,9 +2574,240 @@
                     }
                 }
             }
+        },
+        "/addresses": {
+            "get": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses",
+                "description": "Fetch all addresses/geofences for the organization. An address contains either a circle or polygon geofence describing the address boundaries.",
+                "operationId": "GetOrganizationAddresses",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "List of addresses/geofences. For each address entry, the geofence field will only be populated with either 'circle' or 'polygon', depending on the geofence type.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/Address"
+                            }
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "post": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses",
+                "description": "Add one or more addresses to the organization",
+                "operationId": "AddOrganizationAddresses",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/addressAddParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Addresses were successfully added. No response body is returned."
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/addresses/{addressId}": {
+            "get": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses/{addressId}",
+                "description": "Fetch an address by its id.",
+                "operationId": "GetOrganizationAddress",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "addressId",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the address/geofence",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The address/geofence. The geofence field will only be populated with either 'circle' or 'polygon', depending on the geofence type.",
+                        "schema": {
+                            "$ref": "#/definitions/Address"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses/{addressId}",
+                "description": "Update the name, formatted address, or geofence of an address.",
+                "operationId": "UpdateOrganizationAddress",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/addressUpdateParam" },
+                    {
+                        "name": "addressId",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the address/geofence",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Address was successfully updated. No response body is returned."
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "delete": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses/{addressId}",
+                "description": "Delete an address.",
+                "operationId": "DeleteOrganizationAddress",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "addressId",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the address/geofence",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Address was successfully deleted. No response body is returned."
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
         }
     },
     "definitions": {
+        "Address": {
+            "type": "object",
+            "description": "Information about an address/geofence. Geofences are either a circle or a polygon.",
+            "properties": {
+                "id": {
+                    "type": "integer",
+                    "description": "ID of the address",
+                    "format": "int64",
+                    "example": 123
+                },
+                "name": {
+                    "type": "string",
+                    "description": "Name of the address or geofence",
+                    "example": "Samsara HQ"
+                },
+                "formattedAddress": {
+                    "type": "string",
+                    "description": "The full address associated with this address/geofence, as it might be recognized by maps.google.com",
+                    "example": "350 Rhode Island St, San Francisco, CA"
+                },
+                "geofence": {
+                    "$ref": "#/definitions/AddressGeofence"
+                }
+            }
+        },
+        "AddressGeofence": {
+            "type": "object",
+            "description": "The geofence that defines this address and its bounds. This can either be a circle, or a polygon - only one key should be provided, depending on the geofence type.",
+            "properties": {
+                "circle": {
+                    "type": "object",
+                    "description": "Information about a circular geofence. This field is only populated if the geofence is a circle.",
+                    "properties": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
                 }
             }
+        },
+        "/addresses": {
+            "get": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses",
+                "description": "Fetch all addresses/geofences for the organization",
```
silverspace: The term `addresses/geofences` is a bit confusing. Could we simplify this to just be `addresses`, with maybe a description that an address contains either a circle or polygon geofence describing the address entry?

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "List of addresses/geofences. For each address entry, the geofence field will only be populated with either 'circle' or 'polygon', depending on the geofence type.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#definitions/Address"
```
silverspace: I believe this should be `#/definitions/Address`. Is this rendering correctly on editor.swagger.io?

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                        }
+                    }
+                }
+            },
+            "post": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses",
+                "description": "Add one or more addresses to the organization",
```
silverspace: I believe this endpoint only allows creating a single address entry.

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                "operationId": "AddOrganizationAddresses",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/addressAddParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Addresses were successfully added. No response body is returned."
```
silverspace: We should return the ID of the newly created address (or preferably the entire address payload).

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/addresses/{addressId}",
+                "description": "Update the values of an address",
```
silverspace: `values` is a bit ambiguous. We should list out the fields that can be updated.

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the address/geofence",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Address was successfully updated. No response body is returned."
```
silverspace: Again, this should return the newly modified address entry.

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                        }
+                    }
+                },
+                "polygon": {
+                    "type": "object",
+                    "description": "Information about a polygon geofence. This field is only populated if the geofence is a polygon.",
+                    "properties": {
+                        "vertices": {
+                            "type": "array",
+                            "description": "The vertices of the polygon geofence. These geofence vertices describe the perimeter of the polygon, and must consist of at least 3 vertices and less than 40.",
```
silverspace: This should also mention the order that the vertices should be added (e.g. clockwise).

swagger.json
```diff
@@ -5548,6 +5773,59 @@
             "in": "query",
             "type": "string"
         },
+        "addressAddParam": {
+            "name": "addresses",
```
silverspace: Should `addresses` be plural?

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+            }
+        },
+        "AddressGeofence": {
+            "type": "object",
+            "description": "The geofence that defines this address and its bounds. This can either be a circle, or a polygon - only one key should be provided, depending on the geofence type.",
+            "properties": {
+                "circle": {
+                    "type": "object",
+                    "description": "Information about a circular geofence. This field is only populated if the geofence is a circle.",
+                    "properties": {
```
hchung: shouldn't we include radius as a property?

swagger.json
```diff
@@ -5548,6 +5773,59 @@
...
+                }
+            }
+        },
+        "addressUpdateParam": {
+            "name": "address",
+            "description": "Update parts of an address's value. If the geofence 'circle' or 'polygon' key is specified, the update will change the type of geofence accordingly.",
+            "required": true,
+            "in": "body",
+            "schema": {
+                "type": "object",
```
silverspace: Nit: This shares the same schema as above, with the exception of the requires parameters.

swagger.json
silverspace: Yes, this should include `radiusMeters`.

ehzhang: :facepalm:

swagger.json
ehzhang: As described in the spec, this takes in an array of addresses (so they could add one with an array of size one)

swagger.json
ehzhang: Is there a guideline on what we expect for write action responses? I chose not to return much information to replicate the current responses of our models/controllers code. 
As a side note, I wish that this kind of comment come earlier in the code review or api spec review - I don’t think I did a good job of conveying expectations for the post/patch responses :(

swagger.json
ehzhang: Our code iiuc doesn’t care if it is clockwise or not (you can easily add a geofence in a counter-clockwise manner)
Maybe wed want to specify for clarity?

swagger.json
ehzhang: The schema of the update vs the add seemed substantially different in that effectively every field is optional in the update. Wasn’t sure if there was a cleaner way to handle that in swagger

silverspace: I believe you can reference definitions and still specify a different set of `required` attributes. In any case, this is fine as it is, just means that we'll need to remember to update two call sites if we change the request/response structure.

swagger.json
silverspace: Ok, please ignore this comment then :)

swagger.json
silverspace: Agreed, I should have caught this during the implementation code review. From a customer's perspective, they would need the newly created address id in order to do anything useful with the new address.

silverspace: We can always fix this in a follow-up PR.

swagger.json
silverspace: I thought the clockwise format was the format that Google requires? It's possible that Google Maps sorts the vertices before rendering the polygon, but I'm not sure.

swagger.json
dsharrison: This looks to be rendering fine, but I'll update it anyways.

dsharrison: Nevermind, it was taking the errors a minute to load at the top. I see the error message for this.

swagger.json
```diff
@@ -2574,9 +2574,234 @@
...
+                            "type": "number",
+                            "format": "float",
+                            "description": "The latitude of the center of the circular geofence",
+                            "example": "37.765363"
+                        },
+                        "longitude": {
+                            "type": "number",
+                            "format": "float",
+                            "description": "The longitude of the center of the circular geofence",
+                            "example": "-122.403098"
```
dsharrison: @silverspace It looks like the example gets rendered as-is (without checking the type). Should the type for this example be a float instead of a string?

swagger.json
dsharrison: It looks like we allow multiple addresses in a single `POST` call and the API expects a plural arg.

swagger.json
silverspace: That's a good catch, this should be a float value.

dsharrison: I'll make this update!

swagger.json
dsharrison: I'll add this 😁

swagger.json
dsharrison: I agree that addresses/geofences is a bit confusing, but we use this terminology to refer to addresses in the dashboard so customers will likely be expecting this same terminology in the API. I think we should change this to refer to just addresses but will need to change it everywhere, so I propose we have the API docs match the dashboard for now.

cc @samsarahq/design

swagger.json
dsharrison: Added!

dsharrison:
@silverspace and @hchung this should be ready for another pass!

swagger.json
```diff
@@ -2574,9 +2574,240 @@
...
+                ],
+                "summary": "/addresses",
+                "description": "Fetch all addresses/geofences for the organization. An address contains either a circle or polygon geofence describing the address boundaries.",
+                "operationId": "GetOrganizationAddresses",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "List of addresses/geofences. For each address entry, the geofence field will only be populated with either 'circle' or 'polygon', depending on the geofence type.",
```
silverspace: The phrase `List of addresses/geofences` confuses me. Are the two separate entities? Could we simplify this to `List of addresses for the organization`?

swagger.json
silverspace: This PR is about documenting the code, so we don't need to address here. However, we should definitely start to return the created object in the response payload. Otherwise this API is a black hole and the customer needs to search for the created address id.

swagger.json
silverspace: Could we list out the fields that are eligible to be patched?

swagger.json
```diff
@@ -2574,9 +2574,240 @@
...
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
```
silverspace: Does the code also support a PUT method?

swagger.json
silverspace: Are the first and last vertices required to be the same?

swagger.json
```diff
@@ -2574,9 +2574,240 @@
...
+                },
+                "geofence": {
+                    "$ref": "#/definitions/AddressGeofence"
+                }
+            }
+        },
+        "AddressGeofence": {
+            "type": "object",
+            "description": "The geofence that defines this address and its bounds. This can either be a circle, or a polygon - only one key should be provided, depending on the geofence type.",
+            "properties": {
```
silverspace: Note for future updates (not required for this PR): We should create an additional field specifying the `geofenceType`. OpenAPI 3.0 supports a `discriminator` that can be used to which optional field is populated in a `oneOf` object, and generated clients can automatically make use of this discriminator value.

swagger.json
dsharrison: Let me check!

swagger.json
dsharrison: Sorry I think this reply got lost after the rebase:

I agree that addresses/geofences is a bit confusing, but we use this terminology to refer to addresses in the dashboard so customers will likely be expecting this same terminology in the API. I think we should change this to refer to just addresses but will need to change it everywhere, so I propose we have the API docs match the dashboard for now.

cc @samsarahq/design

swagger.json
dsharrison: Oh neat!

swagger.json
dsharrison: No, we only support a patch.

swagger.json
dsharrison: Sure, should that just take the form of a comma separated list in the description? It doesn't look like there is an option to use markdown or any other type of formatted list.

swagger.json
dsharrison: Absolutely. Let's update the "Go ❤️ APIs @ 🦉" doc to have some basic best practices for each http verb.

ehzhang: Since this would likely map from the mutation response, should that necessarily be an API design principle or a general graphql best practice?

swagger.json
ehzhang: I tried that earlier, but it didn’t seem to a work. It could also be that I didn’t format my fields correctly

swagger.json
dsharrison: I pulled my org's addresses down from the API and it all vertices of the polygon for my drawn address are unique.

dsharrison: Oh god, these all submitted after the GitHub outage was over... wow

PR #65, opened 2018-10-24
swagger: address POST docs: named list addresses required

EmmaOnThursday:
Update address POST docs: addresses being added must be nested within a list called `addresses`. 

image


swagger.json
```diff
@@ -5781,29 +5781,37 @@
         },
         "addressAddParam": {
             "name": "addresses",
-            "description": "Addresses/Geofences to add. Geofences can be circular or a polygon. \nFor each address, only one of 'circle' or 'polygon' should be provided. If both are provided, the geofence will be saved as a polygon.",
+            "description": "List of Addresses/Geofences to add. Geofences can be circular or a polygon. \nFor each address, only one of 'circle' or 'polygon' should be provided. If both are provided, the geofence will be saved as a polygon.",
```
AChelikani: Actually could we decapitalize addresses/geofences here?

swagger.json
```diff
@@ -5781,29 +5781,37 @@
...
             "required": true,
             "in": "body",
             "schema": {
                 "type": "object",
-                "required": [
```
silverspace: We should annotate `addresses` as being required.

PR #68, opened 2018-10-30
Dsharrison/fuel percent

dsharrison:
Add fuel list percent to the Vehicle object for `/fleet/list` (https://github.com/samsara-dev/backend/pull/15341). This PR also corrects the numeric examples on the Vehicle object.

screen shot 2018-10-30 at 10 25 53 am

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2934,6 +2934,12 @@
                     "format": "int64",
                     "description": "Total engine hours for the vehicle.",
                     "example": "1500"
+                },
+                "fuelLevelPercent": {
+                    "type": "float",
+                    "format": "float64",
+                    "description": "The fuel level of the vehicle as a percentage. (0.0 to 1.0)",
+                    "example": 0.3
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -2927,13 +2927,19 @@
...
                     "format": "int64",
                     "description": "Total engine hours for the vehicle.",
-                    "example": "1500"
+                    "example": 1500
+                },
+                "fuelLevelPercent": {
+                    "type": "float",
+                    "format": "float64",
+                    "description": "The fuel level of the vehicle as a percentage. (0.0 to 1.0)",
+                    "example": 0.3
```
silverspace: I forgot to mention this in the other PR -- do we want to restrict the precision of this percentage value?

dsharrison: Probably, this one happens to be an `int64` on the backend so we are ok, but I think providing 2 decimal places would be a good best practice. 

On a related note, golang doesn't have a `round()` or `toPrecision()`? 😕

PR #70, opened 2018-11-02
Add documentation for unassigned segments endpoints

faizandfurious:
This adds documentation for `/fleet/unassigned_driving_segments`, `/fleet/unassigned_driving_segments/annotate`, and `/fleet/unassigned_driving_segments/assign`.

Tested @ http://editor.swagger.io/

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -959,6 +959,176 @@
                 }
             }
         },
+        "/fleet/vehicles/{vehicle_id}/unassigned_driving_segments": {
+            "get": {
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/vehicles/{vehicle_id:[0-9]+}/unassigned_driving_segments",
+                "description": "Get the unassigned driving segments for a specified range.",
+                "operationId": "getUnassignedDrivingSegmentsByVehicleId",
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "name": "vehicle_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the vehicle with the associated unassigned driving segments.",
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    {
+                        "name": "start_ms",
+                        "in": "query",
+                        "required": true,
+                        "description": "Beginning of the time range, specified in milliseconds UNIX time.",
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    {
+                        "name": "end_ms",
+                        "in": "query",
+                        "required": true,
+                        "description": "End of the time range, specified in milliseconds UNIX time.",
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Unassigned driving segments for provided duration and vehicle.",
+                        "schema": {
+                            "type": "object",
+                            "properties": {
+                                "unassigned_driving_segments": {
+                                    "type": "array",
+                                    "items": {
+                                        "$ref": "#/definitions/UnassignedDrivingSegment"
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/fleet/vehicles/{vehicle_id}/unassigned_driving_segments/{segment_id}": {
+            "get": {
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/vehicles/{vehicle_id:[0-9]+}/unassigned_driving_segments/{segment_id}",
+                "description": "Fetch an unassigned driving segment by vehicle ID and segment ID.",
+                "operationId": "getUnassignedDrivingSegmentByVehicleIdAndSegmentId",
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "name": "vehicle_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the vehicle with the associated unassigned driving segments.",
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    {
+                        "name": "segment_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the unassigned driving segment.",
+                        "type": "string",
+                        "format": "string"
+                    },
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Unassigned driving segment log for the specified vehicle ID and segment ID.",
+                        "schema": {
+                            "$ref": "#/definitions/UnassignedDrivingSegment"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/vehicles/{vehicle_id:[0-9]+}/unassigned_driving_segments/{segment_id}",
+                "description": "Update an unassigned driving segment.",
+                "operationId": "patchUnassignedDrivingSegment",
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "name": "vehicle_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the vehicle with the associated unassigned driving segments.",
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    {
+                        "name": "segment_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the unassigned driving segment.",
+                        "type": "string",
+                        "format": "string"
+                    },
+                    {
+                        "name": "unassignedDrivingSegmentParams",
+                        "required": true,
+                        "in": "body",
+                        "schema": {
+                            "type": "object",
+                            "properties": {
+                                "annotation": {
+                                    "type": "string",
+                                    "description": "Annotation to give this segment.",
+                                    "example": "Drive time"
+                                },
+                                "driverId": {
+                                    "type": "integer",
+                                    "format": "int64",
+                                    "description": "Driver ID to assign to this segment.",
+                                    "example": 719
+                                }
+                            }
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The unassigned driving segment that was updated.",
+                        "schema": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -959,6 +959,203 @@
...
+                "description": "Get the unassigned driving segments for a specified range.",
+                "operationId": "get_fleet_unassigned_driving_segments",
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "name": "unassignedDrivingSegmentsParam",
+                        "required": true,
+                        "in": "body",
```
dsharrison: We should use query parameters for these options and should prefer `start-ms` and `duration-ms`.

swagger.json
```diff
@@ -959,6 +959,203 @@
...
+                }
+            }
+        },
+        "/fleet/unassigned_driving_segments/annotate": {
+            "post": {
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/unassigned_driving_segments/annotate",
+                "description": "Annotate an unassigned driving segment.",
```
dsharrison: Are we requiring only one of `start_ms` or `created_at_ms`? Let's add to the description what the options are for identifying the segment to annotate.

swagger.json
```diff
@@ -959,6 +959,203 @@
...
+                                },
+                                "created_at_ms": {
+                                    "type": "integer",
+                                    "format": "int64",
+                                    "description": "Milliseconds UNIX time when this segment was created.",
+                                    "example": 1523059200000
+                                },
+                                "annotation": {
+                                    "type": "string",
+                                    "description": "Annotation to give this segment",
```
dsharrison: Do annotations have a max length?

dsharrison: Super nit: period at the end of this sentence.

swagger.json
```diff
@@ -959,6 +959,203 @@
...
+                        "name": "assignUnassignedDrivingSegmentParams",
+                        "required": true,
+                        "in": "body",
+                        "schema": {
+                            "type": "object",
+                            "required": [
+                                "vehicle_id",
+                                "start_ms",
+                                "created_at_ms",
+                                "driver_id",
```
dsharrison: We should use `camelCase` for request body params.

swagger.json
```diff
@@ -959,6 +959,203 @@
...
+                                    "description": "Driver ID to assign to this segment.",
+                                    "example": 719
+                                }
+                            }
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The unassigned driving segment that was assigned.",
```
dsharrison: Is there some complexity here around what "assigned" means? Does the driver need to accept this change before it is final?

swagger.json
```diff
@@ -3106,6 +3303,42 @@
                 }
             }
         },
+        "UnassignedDrivingSegment": {
+            "type": "object",
+            "description": "A driving segment that has no associated driver",
```
dsharrison: Super nit: period at the end of this sentence.

swagger.json
faizandfurious: @dsharrison I don't immediately see any query parameters that use skeleton case. Want to confirm that we're migrating to that paradigm?

dsharrison: Correct, it looks like that change is still pending from the best practices. Let's stick with the previous best practice here unless @silverspace has more input.

silverspace: That's correct, we are not yet using spinal-case anywhere in our code base. When I was putting together our best practices doc, I saw that most REST guides advised using spinal-case for query parameters, and either camelCase or snake_case for JSON bodies. Since we use Golang, which defaults to camelCase, I've advocated for using camelCase in our JSON request/response bodies.

silverspace: With that said, I don't think we will preserve these existing endpoints in their current form for V2 of our APIs, so we can use an interim naming convention for the legacy APIs that aren't in our Golden API list. I am mostly concerned about the cohesive naming convention in all of our "v2" APIs.

faizandfurious: If we migrate to skeleton case, should we also update the access token parameter here to a skeleton cased version? Otherwise as it stands the parameters are `access_token`, `start-ms`, `end-ms`

faizandfurious:
@dsharrison updated based on comments

swagger.json
```diff
@@ -959,6 +959,182 @@
                 }
             }
         },
+        "/fleet/unassigned_driving_segments": {
```
dsharrison: I talked quickly with @silverspace and I think we want to have this GET for the list, and then use `startMs` and `vehicleId` to create a single `PATCH` for a segment to change the annotation or pending driver assignment. I'll set up a time for us to quickly review.

silverspace: I spoke with Derrek about this endpoint. My thoughts are:
 * We should not use verbs in the URL Path -- we should instead use HTTP methods to represent verbs. 
 * Since unassigned-driving-segments is a first class citizen in our system (or at least we want it to be), this makes sense as a top level entity. 
   * However, we should also support indexing into a resource of this collection using the resource ID, such as `/fleet/unassigned-driving-segments/:id`. 
   * We should support `annotations` and `pendingDriverId` assignments via `PATCH /fleet/unassigned-driving-segments/:id`.

swagger.json
silverspace: That's a good question, our APIs should definitely be internally consistent. We have two options:
1. Update the auth middleware to also accept `access-token`
2. Worry about spinal-case only for newer V2 APIs, which are part of https://paper.dropbox.com/doc/Golden-API-Routes--ARHywquzzMP3nl1Lnk8P1DzfAg-A6hbRcUq2NQFDuxtKb0si

I think it's ok to proceed with snake_case query parameters for now, since that's consistent with what all of our other APIs use.

swagger.json
```diff
@@ -959,6 +959,182 @@
...
+            "get": {
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/unassigned_driving_segments",
+                "description": "Get the unassigned driving segments for a specified range.",
+                "operationId": "get_fleet_unassigned_driving_segments",
```
silverspace: Nit: s/get/list

swagger.json
```diff
@@ -959,6 +959,182 @@
                 }
             }
         },
+        "/fleet/unassigned_driving_segments": {
+            "get": {
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/unassigned_driving_segments",
+                "description": "Get the unassigned driving segments for a specified range.",
```
silverspace: `specified time range`

swagger.json
```diff
@@ -959,6 +959,182 @@
...
+                "operationId": "get_fleet_unassigned_driving_segments",
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "name": "start_ms",
+                        "in": "path",
```
silverspace: Should be in `query` instead of `path`

swagger.json
```diff
@@ -959,6 +959,182 @@
...
+                        "required": true,
+                        "description": "Beginning of the time range, specified in milliseconds UNIX time.",
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    {
+                        "name": "end_ms",
+                        "in": "path",
```
silverspace: Same as above, should be a query parameter.

swagger.json
```diff
@@ -959,6 +959,182 @@
...
+                        "required": true,
```
silverspace: Is this required, or do we default to a reasonable value?

swagger.json
```diff
@@ -959,6 +959,182 @@
...
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/fleet/unassigned_driving_segments/annotate": {
```
silverspace: This should be a `PATCH /fleet/unassigned_driving_segments/:id` rather than `PATCH /fleet/unassigned_driving_segments/annotate` (verb should be the HTTP method, not a path parameter).

swagger.json
```diff
@@ -959,6 +959,182 @@
...
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/fleet/unassigned_driving_segments/assign": {
```
silverspace: Same as above, could we change this to `PATCH /fleet/unassigned_driving_segments/:id`?

swagger.json
```diff
@@ -3106,6 +3282,42 @@
                 }
             }
         },
+        "UnassignedDrivingSegment": {
+            "type": "object",
+            "description": "A driving segment that has no associated driver.",
+            "properties": {
```
silverspace: Could we give this unassigned_driving_segment resource a unique identifier, so it can be accessed via `/fleet/unassigned_driving_segments/:id`?

swagger.json
faizandfurious: So would that make these the same endpoints? Or do you mean
`/fleet/unassigned_driving_segments/:id/assign`

swagger.json
faizandfurious: Similar as the other comment, do you mean
`/fleet/unassigned_driving_segments/:id/annotate`

silverspace: No, I think the annotation is as simple as PATCH updating the `annotation` field of the unassigned driving segment. Is the annotation a simple string attribute of the driving segment, or a timeseries value?

swagger.json
silverspace: I was thinking we `PATCH /fleet/unassigned_driving_segments/:id`, and the `body` JSON includes the new `pending_driver_id` value.

swagger.json
```diff
@@ -3106,6 +3276,42 @@
                 }
             }
         },
+        "UnassignedDrivingSegment": {
```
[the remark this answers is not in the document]
faizandfurious: Great catch, let me add that to the data structure.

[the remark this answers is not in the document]
faizandfurious: Added to the data structure, including an base64'd epoch milliseconds example

swagger.json
```diff
@@ -3106,6 +3276,47 @@
...
+            "type": "object",
+            "description": "A driving segment that has no associated driver.",
+            "properties": {
+                "segment_id": {
```
silverspace: Should this be `id` or `segment_id`?

swagger.json
```diff
@@ -959,6 +959,176 @@
                 }
             }
         },
+        "/fleet/vehicles/{vehicle_id}/unassigned_driving_segments": {
```
silverspace: Is unassigned driving segment a concept specific to HOS? Should we put this under an HOS namespace?

faizandfurious: I think the current namespace is ok

swagger.json
faizandfurious: Hey, good point, I'll change this to `id` on the `UnassignedDrivingSegment` object.

faizandfurious: done

PR #72, opened 2018-11-06
Add pendingOnDutyMs and pendingDriveMs to fleet/drivers/hos_daily_logs

faizandfurious:


swagger.json
```diff
@@ -3618,6 +3618,18 @@
                     "description": "Duration in milliseconds that driver was driving during the requested time range.",
                     "example": 21600000
                   },
+                  "pendingOnDutyMs": {
+                    "type": "number",
+                    "format": "int64",
+                    "description": "Duration in milliseconds that driver was on duty during the requested time range for pending logs.",
```
dsharrison: Nit: do we want to rephrase this to:

`Duration in milliseconds that driver was on duty during the requested time range if all pending logs were to become final.`

PR #74, opened 2018-11-07
swagger: API spec for dvir POST endpoint

EmmaOnThursday:
A DVIR POST endpoint matching these specs is in development. 

This POST will, for now, only allow "mechanic-style" DVIRs rather than full driver DVIRs. Users will not be able to specify specific defects that need attention, but will be able to mark a vehicle or trailer safe or unsafe. 
Marking a vehicle or trailer as safe resolves any previous unsafe DVIRs. 

A POST to this endpoint returns the newly-created DVIR.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1013,7 +1013,7 @@
                     "Fleet","Default"
                 ],
                 "summary": "/fleet/maintenance/dvirs",
-                "description": "Get the DVIR for the org with the time constraints",
+                "description": "Get DVIRs for the org within provided time constraints",
                 "operationId": "get_dvirs",
                 "parameters": [
                     {
@@ -1043,9 +1043,35 @@
                 ],
                 "responses": {
                     "200": {
-                        "description": "DVIR's for the specified group and duration.",
+                        "description": "DVIRs for the specified duration.",
                         "schema": {
                             "$ref": "#/definitions/DvirListResponse"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "post": {
+                "tags": [
+                    "Fleet","Default"
+                ],
+                "summary": "/fleet/maintenance/dvirs",
+                "description": "Create a new dvir, marking a vehicle or trailer safe or unsafe.",
+                "operationId": "create_dvir",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/createDvirParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Newly created DVIR.",
+                        "schema": {
+                            "$ref": "#/definitions/DvirBase"
                         }
                     },
                     "default": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -5954,6 +5984,55 @@
...
+                        "description": "Whether any previous defects were corrected.",
+                        "example": true
+                    },
+                    "previousDefectsIgnored": {
+                        "type": "boolean",
+                        "description": "Whether any previous defects were ignored.",
+                        "example": false
+                    },
+                    "odometerMiles": {
+                        "type": "integer",
```
silverspace: Is this an integer or a floating point value?

swagger.json
```diff
@@ -5954,6 +5984,55 @@
...
+            "description": "DVIR creation body",
+            "required": true,
+            "in": "body",
+            "schema": {
+                "type": "object",
+                "required": [
+                    "vehicleIsSafe"
+                ],
+                "properties": {
+                    "vehicleIsSafe" : {
```
silverspace: Would it make sense to split this DVIR model into a separate definition, and reference the DVIR object from `createDvirParam`?

swagger.json
```diff
@@ -1055,6 +1055,32 @@
                         }
                     }
                 }
+            },
+            "post": {
+                "tags": [
+                    "Fleet","Default"
+                ],
+                "summary": "/fleet/maintenance/dvirs",
+                "description": "Create a minimal new dvir, marking a vehicle safe or unsafe.",
```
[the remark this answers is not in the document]
silverspace: Should we describe that mechanics are not required to submit all DVIR related fields?

swagger.json
EmmaOnThursday: All our other odometer numbers are integers, and the DVIR list endpoint returns odometers as integers - I'd like to keep that consistent.

swagger.json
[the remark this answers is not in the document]
EmmaOnThursday: Rewording to `Create a new dvir, marking a vehicle or trailer safe or unsafe.` 
Do we normally specify which fields are/aren't required in descriptions? I thought swagger's layout handled that.

swagger.json
```diff
@@ -1043,7 +1043,7 @@
                 ],
                 "responses": {
                     "200": {
-                        "description": "DVIR's for the specified group and duration.",
+                        "description": "DVIRs for the specified group and duration.",
```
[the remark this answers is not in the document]
EmmaOnThursday: DVIRs are group-scoped, not org-scoped. `groupId` must be included for multi-group orgs.

swagger.json
EmmaOnThursday: I don't see a benefit, since this is only used here, but if that's our convention I don't object!

EmmaOnThursday: DVIR returned from the list endpoint, and the response from this POST, share a struct: `DvirBase`.

swagger.json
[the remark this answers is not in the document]
EmmaOnThursday: Ah yes, drivers have one field that's required that mechanics don't have - specifying if a DVIR is pre- or post-trip. That field isn't included yet though, since mechanics don't use it.

swagger.json
silverspace: I see, I think users of this API may be confused why the POST doesn't accept fields from the `DvirBase` model, such as `vehicleDefects`. Would it make sense to have this endpoint be a fully-fledged DVIR creation endpoint, such that it takes a superset of `DvirBase` (minus the DVIR `id` value).

davidcmoody: From a customer viewpoint, it would be nice to be able to generate full-fledged DVIRs.

PR #76, opened 2018-11-09
apidocs: add proposed external id changes for arcbest

AChelikani:
Proposed changes:

1. Update driver creation payload to allow a passed in `external_id` key and value.

Example: 
```
{
    ...
    "driver": {
        "driverId": 1,
        "externalIds": {
             "payrollId": "123",
             "externalDriverId": "123456"
         }
    }
}
```

2. Update all endpoints that previously routed based on `driver_id` or `vehicle_id` to now also allow the `external_id` in the following format (note that old format is still allowed as well):

Example: `GET /fleet/drivers/arcbestDriverId:123456`
Example: `GET /fleet/drivers/1`

Example: `POST /fleet/vehicles/arcbestVehicleId:ABFS18600/dispatch/routes`
Example: `POST /fleet/vehicles/1/dispatch/routes`

3. Add a patch endpoint for drivers and vehicles that allows updating a driver or vehicle based on either `driver_id`, `vehicle_id`, or `external_id`. The intension is to allow for retroactively assigning or modifying an `external_id` to a driver or vehicle. 

Example:
```
PATCH /fleet/vehicles/1
PAYLOAD
{
    "vehicle": {
        "vehicleId": 1,
        "externalIds": {
             "externalVehicleId": "CITY1234"
         }
    }
    ...
}
```

screen shot 2018-11-12 at 11 35 28 am

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -568,25 +568,24 @@
                 }
             }
         },
-        "/fleet/drivers/{driver_id}": {
-            "parameters": [
+        "/fleet/drivers/{driver_id | external_id}": {
+            "get": {
+              "parameters": [
                 {
                     "$ref": "#/parameters/accessTokenParam"
                 },
                 {
-                    "name": "driver_id",
+                    "name": "driver_id | external_id",
                     "in": "path",
                     "required": true,
-                    "description": "ID of the driver.",
-                    "type": "integer",
-                    "format": "int64"
+                    "description": "Either the ID of the driver (integer) or corresponding external ID (alphanumeric string). One of these two parameters must be specified to identify the driver.",
+                    "type": "string"
                 }
             ],
-            "get": {
                 "tags": [
                     "Default", "Fleet", "Drivers"
                 ],
-                "summary": "/fleet/drivers/{driver_id:[0-9]+}",
+                "summary": "/fleet/drivers/{driver_id:[0-9]+ | external_id:[a-zA-Z0-9]+}",
                 "description": "Fetch driver by id.",
                 "operationId": "getDriverById",
                 "consumes": [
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -568,25 +568,24 @@
                 }
             }
         },
-        "/fleet/drivers/{driver_id}": {
-            "parameters": [
+        "/fleet/drivers/{driver_id OR external_id}": {
```
silverspace: Would we make this `/fleet/drivers/{driver_id|external_id}`?

swagger.json
```diff
@@ -568,25 +568,24 @@
...
+            "get": {
+              "parameters": [
                 {
                     "$ref": "#/parameters/accessTokenParam"
                 },
                 {
-                    "name": "driver_id",
+                    "name": "driver_id OR external_id",
```
silverspace: Same as above, could we make this `driver_id|external_id`

swagger.json
```diff
@@ -568,25 +568,24 @@
...
                     "in": "path",
                     "required": true,
-                    "description": "ID of the driver.",
-                    "type": "integer",
-                    "format": "int64"
+                    "description": "ID of the driver (integer) or corresponding external ID (alphanumeric string).",
```
silverspace: We may want to provide a more detailed description of this parameter, since it is really a `oneOf(driverId, externalId)`.

swagger.json
```diff
@@ -568,25 +568,24 @@
...
+                    "type": "string"
                 }
             ],
-            "get": {
                 "tags": [
                     "Default", "Fleet", "Drivers"
                 ],
-                "summary": "/fleet/drivers/{driver_id:[0-9]+}",
+                "summary": "/fleet/drivers/{driver_id:[0-9]+ OR external_id:[a-zA-Z0-9]+}",
```
silverspace: Nit: `/fleet/drivers/{(driver_id:[0-9]+)|(external_id:[a-zA-Z0-9]+)}`

swagger.json
silverspace: Could we move this to the common `#/parameters/` section, so this can be re-used across HTTP methods?

swagger.json
```diff
@@ -3429,7 +3519,12 @@
                     "format": "int64",
                     "description": "ID of the group if the organization has multiple groups (uncommon).",
                     "example": 101
-                }
+                },
+                "arcbestDriverId": {
```
silverspace: This should be inside an `externalIds` dictionary. Could we instead use:

```
     "externalIds": {
                    "type": "object",
                    "additionalProperties": {
                      "type": "string"
                    },
                    "example": {"scac": "city1234", "myid": "foo234"}
      },
```

swagger.json
```diff
@@ -4617,12 +4717,22 @@
                             "description":  "ID of the driver assigned to the dispatch job.",
                             "example": 444
                         },
+                        "arcbest_driver_id": {
```
silverspace: Same as above, this should be a dictionary of key=value pairs, rather than a hard coded `arcbest_driver_id` key.

swagger.json
```diff
@@ -5947,6 +6067,73 @@
...
+        "driverParam" : {
+          "name": "driverParam",
+          "description": "Driver update body",
+          "required": true,
+          "in": "body",
+          "schema": {
+            "$ref": "#/definitions/Driver"
+          }
+        },  
+        "createDvirParam": {
```
silverspace: Why the `createDvirParam` change?

swagger.json
```diff
@@ -4665,17 +4792,44 @@
...
                             "example": 55
                         },
-                        "driver_id": {
-                            "type": "integer",
-                            "format": "int64",
-                            "description":  "ID of the driver assigned to the dispatch job.",
-                            "example": 444
+                        "driver": {
+                            "type": "object",
+                            "description": "Driver object",
```
silverspace: Could this driver object reference a `#/definitions/ExternalDriver`, so it can be shared among all of the response payloads?

swagger.json
```diff
@@ -4738,17 +4892,44 @@
...
             ],
             "properties": {
-                "vehicle_id": {
-                    "type": "integer",
-                    "format": "int64",
-                    "description":  "ID of the vehicle assigned to the dispatch route. Note that vehicle_id and driver_id are mutually exclusive. If neither is specified, then the route is unassigned.",
-                    "example": 444
+                "vehicle": {
+                    "type": "object",
+                    "description": "Vehicle object",
```
silverspace: Same as above, could we define this in an `#/definitions/ExternalVehicle` object?

PR #77, opened 2018-11-10
Dsharrison/users crud api

dsharrison:
Add documentation for the future user CRUD API. This is heavily modeled on the user csv upload functionality available today.

https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/dsharrison/users-crud-api/swagger.json#operation/get_users

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -40,9 +40,202 @@
         {
             "name": "Default",
             "description": "All Samsara API endpoints"
+        },
+        {
+            "name": "Users",
+            "description": "Access to user data"
         }
     ],
     "paths": {
+        "/users": {
+            "get": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users",
+                "description": "Get all users in the organization.",
+                "operationId": "list_users",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "List of users.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/User"
+                            }
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "post": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users",
+                "description": "Add a user to the organization.",
+                "operationId": "create_user",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/createUserParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "User successfully created.",
+                        "schema": {
+                            "$ref": "#/definitions/User"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/users/{user_id}": {
+            "get": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users/{user_id:[0-9]+}",
+                "description": "Get a user.",
+                "operationId": "get_user",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "user_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the user.",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "User record.",
+                        "schema": {
+                            "$ref": "#/definitions/User"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "delete": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users/{user_id:[0-9]+}",
+                "description": "Remove a user from the organization.",
+                "operationId": "delete_user",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "user_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the user.",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The user was successfully removed."
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users/{user_id:[0-9]+}",
+                "description": "Update some fields on a user. Only fields to be changed need to be supplied. Note that if you change an object or collection, you must supply the complete collection. For example, to add a tag role to a user, you must specify all tag roles that the user should have.",
+                "operationId": "update_user",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/updateUserParam" },
+                    {
+                        "name": "user_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the user.",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "User successfully updated.",
+                        "schema": {
+                            "$ref": "#/definitions/User"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "put": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users/{user_id:[0-9]+}",
+                "description": "Replace all fields on a user with the supplied fields. Any fields not supplied will be set to null.",
+                "operationId": "replace_user",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    { "$ref": "#/parameters/createUserParam" },
+                    {
+                        "name": "user_id",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the user.",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "User successfully created.",
+                        "schema": {
+                            "$ref": "#/definitions/User"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
         "/sensors/list": {
             "post": {
                 "tags": [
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -6257,6 +6474,117 @@
...
+                        "description": "The name of the role the user is assigned to for the organization. Leave this blank to give only tag access. This must match one of the preset roles or the name of a custom role in the organization. This value is case insensitive.",
+                        "example": "Read-Only Admin"
+                    },
+                    "tagRoles" : {
+                        "type": "array",
+                        "description": "If specified, the user will be a tag administrator for these tags. If left blank, the user has access to the entire organization.",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "tagName": {
```
dsharrison: We allow tags to be referenced by name during csv upload so I am proposing we do the same here.

swagger.json
```diff
@@ -6327,6 +6327,117 @@
...
+                                },
+                                "role" : {
+                                    "type": "string",
+                                    "description": "The name of the role the user is assigned to for this tag. This must match one of the preset roles or the name of a custom role in the organization. This value is case insensitive.",
+                                    "example": "Read-Only Admin"
+                                }
+                            }
+                        }
+                    },
+                    "authType": {
```
AChelikani: Should we stick `authType` next to `name` and `email` to keep all the required params together?

swagger.json
```diff
@@ -5850,6 +5850,72 @@
...
+                        "$ref": "#/definitions/UserTagRole"
+                    }
+                },
+                "authType": {
+                    "type": "string",
+                    "enum": [
+                        "basic",
+                        "saml"
+                    ],
+                    "description": "The authentication type the user uses to authenticate. To use SAML your organization must have a configured SAML integration."
```
AChelikani: Should we change this to `To use SAML this organization must have a configured SAML integration` since we don't use second person anywhere in the doc except in the intro line?

swagger.json
```diff
@@ -40,9 +40,160 @@
...
     ],
     "paths": {
+        "/users": {
+            "get": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users",
+                "description": "Get all users in your organization.",
+                "operationId": "get_users",
```
silverspace: Nit: would we call this `list_users` for consistency with `list collection` vs `get resource`?

swagger.json
```diff
@@ -40,9 +40,160 @@
...
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the user.",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The user was successfully removed."
```
silverspace: Long term I'd like to start returning `204: No Content` for successful DELETE operations that don't return any response.

swagger.json
```diff
@@ -40,9 +40,160 @@
...
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users",
+                "description": "Update a user.",
```
silverspace: Nit: we may want to describe in detail what the PATCH operation does, since many users are probably unaware of the difference between PUT and PATCH.

swagger.json
```diff
@@ -40,9 +40,160 @@
...
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
```
silverspace: Could we also add support for a `put` operation?

swagger.json
```diff
@@ -5850,6 +6001,72 @@
             "items": {
                 "$ref": "#/definitions/DocumentType"
             }
+        },
+        "User": {
+            "type": "object",
+            "properties": {
+                "id": {
+                    "type": "integer",
```
silverspace: `"format": "int64"`

swagger.json
```diff
@@ -5850,6 +6001,72 @@
...
+                    "description": "The id of the User record.",
+                    "example": 123
+                },
+                "name" : {
+                    "type": "string",
+                    "description": "The first and last name of the user.",
+                    "example": "Jane Doe"
+                },
+                "email": {
+                    "type": "string",
```
silverspace: We can define `"format": "email"` as a hint of the format type, even though it isn't supported in swagger: https://swagger.io/docs/specification/data-models/data-types/#string

swagger.json
```diff
@@ -5850,6 +6001,72 @@
...
+                    "description": "The email address of this user.",
+                    "example": "user@company.com"
+                },
+                "organizationRole" : {
+                    "type": "string",
+                    "description": "The name of the role the user is assigned to for the organization. This will be blank for users that only have access to specific tags.",
+                    "example": "Read-Only Admin"
```
silverspace: We should expose this as an enum field: https://swagger.io/docs/specification/2-0/enums/

swagger.json
```diff
@@ -5850,6 +6001,72 @@
...
+                },
+                "tagRoles" : {
+                    "type": "array",
+                    "description": "The specific tags this user has access to. This will be blank for users that have full access to the organization.",
```
silverspace: Is this only available if `organizationRole` is `Tag Admin`? If so, we should call that out (I think we can use organizationRole as a discriminator for this type).

swagger.json
```diff
@@ -5850,6 +6001,72 @@
...
+                        },
+                        "name": {
+                            "type": "string",
+                            "description": "Name of this tag.",
+                            "example": "North Carolina Distribution"
+                        }
+                    }
+                },
+                "role": {
+                    "type": "string",
```
silverspace: Could this be an enum?

swagger.json
```diff
@@ -5850,6 +6001,72 @@
...
+                "tag": {
+                    "type": "object",
+                    "properties": {
+                        "id": {
+                            "type": "integer",
+                            "format": "int64",
+                            "description": "The ID of this tag.",
+                            "example": 12345
+                        },
+                        "name": {
```
silverspace: Should we return a full tag object here? (Including `parentTagId`, etc)

swagger.json
```diff
@@ -6257,6 +6474,117 @@
...
+                        "enum": [
+                            "basic",
+                            "saml"
+                        ],
+                        "description": "The authentication type the user uses to authenticate. To use SAML your organization must have a configured SAML integration."
+                    }
+                }
+            }
+        },
+        "updateUserParam": {
```
silverspace: Nit: Can `updateUserParam` and `createUserParam` share the same base object, using `allOf`, to avoid copy/pasting these fields?

swagger.json
dsharrison: Sure!

swagger.json
dsharrison: Makes sense to me.

swagger.json
dsharrison: In our CSV upload we specify that users should leave the organization role blank to create a tag admin.

swagger.json
dsharrison: I wasn't aware of `allOf`, yes that makes sense!

swagger.json
silverspace: Just thinking out loud -- would it make sense to specify all of the standard role types as an enum, with an additional `CustomRole` type as an option, that acts as a discriminator for another `customRole` field?

swagger.json
```diff
@@ -5850,6 +5850,72 @@
             "items": {
                 "$ref": "#/definitions/DocumentType"
             }
+        },
+        "User": {
+            "type": "object",
+            "properties": {
+                "id": {
+                    "type": "integer",
+                    "description": "The id of the User record.",
```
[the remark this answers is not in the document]
dsharrison: Yes.

swagger.json
dsharrison: I thought about doing that, but I am not sure what a consumer would do with that information. Do you have a preference?

swagger.json
dsharrison: Do you think this would lead to cleaner error handling?

swagger.json
dsharrison: I like this idea, but it throws a ton of errors in the swagger editor preview.

swagger.json
dsharrison: I like this more

swagger.json
dsharrison: Added a more detailed description.

swagger.json
silverspace: I'm primarily thinking about our generated SDKs (which I plan to start generating out soon). If these magic values are encoded in the swagger schema, then the SDKs will have a set of auto-complete options.

swagger.json
silverspace: No preference, I'm primarily concerned that this field is future compatible with the tag object in case we decide to include additional tag fields (e.g. external_ids).

dsharrison: Good point, I will add `parentTagId`.

swagger.json
dsharrison: Do you want that change made here?

swagger.json
```diff
@@ -40,9 +40,202 @@
...
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users/{user_id:[0-9]+}",
+                "description": "Get a user.",
+                "operationId": "get_user",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "user_id",
```
silverspace: Nit: might make sense to move this to `#/parameters/userId` and share it with the other /users endpoints.

swagger.json
```diff
@@ -6374,6 +6683,31 @@
...
+            "required": true,
+            "in": "body",
+            "schema": {
+                "type": "object",
+                "required": [
+                    "name",
+                    "email",
+                    "authType"
+                ],
+                "$ref": "#/definitions/UpsertUserBase"
```
dsharrison: @silverspace When using `$ref` for the schema I am not seeing the required properties being marked in the UI. Is there a different way I should be handling required properties?

swagger.json
```diff
@@ -40,9 +40,202 @@
...
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": [
+                    "Default", "Users"
+                ],
+                "summary": "/users/{user_id:[0-9]+}",
+                "description": "Update some fields on a user. Only fields to be changed need to be supplied. Note that if you change an object or collection, you must supply the complete collection. For example, to add a tag role to a user, you must specify all tag roles that the user should have.",
```
silverspace: 👍

swagger.json
silverspace: 👍

swagger.json
silverspace: I think the issue is that JSON schema treats the `allOf` as a literal concatenation of the values in the array, and the required attributes are not applied to the result. If you update the required attributes in the `#/definitions` section, they should show up as required. For example:

```
        "UserBase": {
            "type": "object",
            "required": [ "name", "email", "authType" ],
```

swagger.json
silverspace: Your call :)

swagger.json
silverspace: This is caused by https://github.com/swagger-api/swagger-editor/issues/1664#issuecomment-372098616

A workaround is to use `allOf` to specify required attributes of a $ref:

```
        "createUserParam": {
            "name": "user",
            "description": "User properties. Only one of organizationRole or tagRoles is required.",
            "required": true,
            "in": "body",
            "schema": {
              "allOf": [
                  {
                     "$ref": "#/definitions/UpsertUserBase"
                  },
                ],
                "type": "object",
                "required": [
                    "name",
                    "email",
                    "authType"
                ],
            }
        },
```

swagger.json
dsharrison: Deferred this.

dsharrison:
Great questions!

For the multi-group org question let's chat with @silverspace, My understanding is that we are going to be offering limited API support for them (hence the removal of the `group_id` param).

This iteration is only allowing for a single user to be created via `POST`. If it is minimal additional effort, we could expand this to allow for multiple groups.

For `PUT` / `PATCH`. This request came from @silverspace so let's chat with him about it.

PR #80, opened 2018-11-14
swagger: API spec for /contacts/ endpoints

sjain28:
Adds documentation for newly added `/contacts/` GET endpoints. Contacts API supports 
* GET organization contacts
* GET contact by id

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2723,6 +2723,71 @@
                     }
                 }
             }
+        },
+        "/contacts": {
+            "get": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/contacts",
+                "description": "Fetch all contacts for the organization.",
+                "operationId": "ListContacts",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "List of contacts.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/Contact"
+                            }
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/contacts/{contact_id}": {
+            "get": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/contacts/{contactId}",
+                "description": "Fetch a contact by its id.",
+                "operationId": "GetOrganizationContact",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" },
+                    {
+                        "name": "contactId",
+                        "in": "path",
+                        "required": true,
+                        "description": "ID of the contact",
+                        "type": "integer",
+                        "format": "int64"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The contact.",
+                        "schema": {
+                            "$ref": "#/definitions/Contact"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
         }
     },
     "definitions": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -5796,6 +5861,55 @@
...
+                },
+                "phone": {
+                    "type": "string",
+                    "description": "Phone number of the contact",
+                    "example": "111-222-3344"
+                },
+                "webhook": {
+                    "type": "string",
+                    "description": "Webhook associated with the contact",
+                    "example": "???"
```
EmmaOnThursday: looks like we need an example here

sjain28: @silverspace any recommendations on whether we should return webhooks here, and what an example webhook looks like?

silverspace: I think this is an unused field that we should remove from the contacts table.

swagger.json
```diff
@@ -2723,6 +2723,71 @@
...
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
+        "/contacts/{contactId}": {
```
dsharrison: Should this be `contact_id`?

swagger.json
```diff
@@ -5796,6 +5861,55 @@
...
+                    "format": "int64",
+                    "example": 123
+                },
+                "userId": {
+                    "type": "integer",
+                    "description": "User ID of the contact",
+                    "format": "int64",
+                    "example": 123
+                },
+                "orgId": {
```
dsharrison: Do we need to return this? The caller should already know the organization.

sjain28: Removed the field.

swagger.json
sjain28: Removed the field.

swagger.json
sjain28: Not sure - for some entities (addresses, drivers) we camelcase the id and for others (assets, vehicles) we snake case it. @silverspace do you have opinions here?

silverspace: Let's `camelCase` JSON request/response bodies, and `spinal-case` query parameters and PATH parameters. https://paper.dropbox.com/doc/API-Best-Practices-HNfOvJK77p8UIC5BUFO8S

silverspace: Sorry I didn't fully read this question -- the name of the route template variable doesn't matter much to me, that's internal to our swagger spec.

silverspace: Let's use `snake_case` since that's what most of the rest of our swagger spec uses, just to be internally consistent.

swagger.json
```diff
@@ -2723,6 +2723,71 @@
...
             }
+        },
+        "/contacts": {
+            "get": {
+                "tags": [
+                    "Default", "Fleet"
+                ],
+                "summary": "/contacts",
+                "description": "Fetch all contacts for the organization.",
+                "operationId": "GetOrganizationContacts",
```
silverspace: Nit: could we cal this `ListContacts`?

swagger.json
```diff
@@ -5796,6 +5861,44 @@
             "items": {
                 "$ref": "#/definitions/DocumentType"
             }
+        },
+        "Contact": {
+            "type": "object",
+            "description": "Information about a contact.",
```
silverspace: Could we describe this as being a notification contact for alerts?

swagger.json
```diff
@@ -5796,6 +5861,44 @@
...
+            "properties": {
+                "id": {
+                    "type": "integer",
+                    "description": "ID of the contact",
+                    "format": "int64",
+                    "example": 123
+                },
+                "userId": {
```
silverspace: What is the `userId` in our system? Contacts don't have references to our `users` table, do they?

sjain28: `contacts` does in fact have a FK to `users`

swagger.json
sjain28: Changed to `contact_id`

swagger.json
sjain28: Done

swagger.json
sjain28: Reworded the description.

swagger.json
sjain28: Oh, interesting. The Model struct has a foreign key annotation. Weird.

PR #84, opened 2018-11-19
swagger: updating address api docs to include notes and contacts

sjain28:
This PR updates address api docs to include notes and contact fields in address query responses and mutation parameters.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2813,6 +2813,15 @@
                 },
                 "geofence": {
                     "$ref": "#/definitions/AddressGeofence"
+                },
+                "contacts": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Contact"
+                    }
+                },
+                "notes": {
+                    "$ref": "#/definitions/AddressNotes"
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -5893,6 +5907,15 @@
                     "example": "111-222-3344"
                 }
             }
+        },
+        "ContactIds": {
+            "type": "array",
+            "description": "A list of IDs for contacts who can be notified about alerts",
```
EmmaOnThursday: this description ties ContactIds exclusively to alerts, but in fact it's just a list of Ids and the type could be reused elsewhere.

sjain28: @silverspace requested in the spec for `Contacts` API that we describe contacts as `Information about a notification contact for alerts.`

EmmaOnThursday: That makes sense in the context of the `Contact` object but this particular object is just a list of IDs. If we want to bring alerts into it can we explicitly tie it back to addresses also? this just looks wrong in cases where someone is just trying to create an address and is like, "hm, what is this list of contacts ids for?".

silverspace: We could update this description when we support more operations on address based contacts? I'm fine with whatever you guys decide for this description (and I'm also OOO this week, so won't be following this thread).

sjain28: I'm ok with rewording this so that it's not tied to alerts!

sjain28: @EmmaOnThursday updated the text here to be more general.

PR #85, opened 2018-11-21
swagger: add tags to addresses

EmmaOnThursday:
Adds tag ID parameters to address POST and PATCH, and adds tags to responses for address single and multiple GET. 

For discussion: 
~Does it make sense to include the entire tag object in the GET responses? a ton of data comes back in the list of tags, because it's returning every entity under every tag. Would it be better to return just the list of tag IDs?~ Outcome: we'll be returning just the name & id of each tag. 

[Sandbox here.](http://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/6ea0894f5a1075a466a42596f39783d642a673a0/swagger.json#operation/AddOrganizationAddresses) If you click into the GET /addresses endpoint you can see an example tags response in context.

To be landed once this functionality is fully implemented.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2667,7 +2667,7 @@
                     "Default", "Fleet"
                 ],
                 "summary": "/addresses/{addressId}",
-                "description": "Update the name, formatted address, or geofence of an address.",
+                "description": "Update the name, formatted address, geofence, notes, or tag and contact Ids for an address. The set of tags or contacts associated with this address will be updated to exactly match the list of IDs passed in. To remove all tags or contacts from an address, pass an empty list; to remove notes, pass an empty string.",
                 "operationId": "UpdateOrganizationAddress",
                 "parameters": [
                     { "$ref": "#/parameters/accessTokenParam" },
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4581,6 +4587,15 @@
                     }
                 }
         },
+        "TagIds": {
+            "type": "array",
+            "description": "A list of tag IDs.",
```
dsharrison: I think we should have done this for contactIds too, but can we describe how this list works on an update? Does the list itself behave like a `PUT` where I need to supply the complete list?

EmmaOnThursday: It does, but that's more a function of the endpoint than the actual object. I can add a note about that to the address PATCH description.

swagger.json
```diff
@@ -4586,6 +4592,35 @@
                     }
                 }
         },
+        "TagMetadata": {
+            "type": "object",
+            "required": [
+                "id",
+                "name"
```
silverspace: This is a little awkward to require the tag name. Are we going to do anything with the tag name? I don't think this should update the existing tag's name. I'd be inclined to require only `id` for the input. In the future we may extend `external ids` to tags, which would allow them to specify `key:value` en lieu of `id`.

PR #90, opened 2018-12-05
Fleet Driver Settings endpoint updates

alberthu16:
Live demo for these changes: https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/albert/arcbest-fleet-driver-updates/swagger.json

These doc updates are meant to reflect the changes specified in this [product spec](https://docs.google.com/document/d/1b9WG8e37rcG2P479nVNJx1D69evjSJRG-Pk9i54KnHA/edit#).

More context here and up-to-date specs for field descriptions here in the [engineering implementation spec](https://paper.dropbox.com/doc/Implementation-Spec-Driver-Settings-API-updates--ATHutWfTJT7~qla1JI2Qq77lAg-b0Y32QZTY8zTB8dCKk3hp#:h2=Use-the-new-graphql-endpoint-w).

Overall, the main change is that we want to expose more fields on the `/fleet/driver` endpoints for creating, reading, and updating driver objects.

There are three big changes:
1) return/accept more driver fields for POST /fleet/drivers/create, PATCH /fleet/drivers/:id, and all GETs
2) for POST /fleet/drivers -> return all driver info for each driver, not just id, name, and external id
3) deprecate old fields such as "eldBigDayExemptionEnabled" by removing them from the documentation here

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -747,7 +747,7 @@
                 ],
                 "responses": {
                     "200": {
-                        "description": "Returns the successfully created the driver.",
+                        "description": "Returns the successfully created driver.",
                         "schema": {
                             "$ref": "#/definitions/Driver"
                         }
--- a/swagger.json
+++ b/swagger.json
@@ -4187,28 +4187,24 @@
                 "drivers": {
                     "type": "array",
                     "items": {
-                        "type": "object",
-                        "properties": {
-                            "id": {
-                                "type": "integer",
-                                "format": "int64",
-                                "description": "ID of the driver.",
-                                "example": 444
+                        "allOf": [
+                            {
+                                "$ref": "#/definitions/Driver"
                             },
-                            "name": {
-                                "type": "string",
-                                "description": "Name of the driver.",
-                                "example": "Fred Jacobs"
-                            },
-                            "externalIds" : {
-                              "type": "object",
-                              "description": "Dictionary of external IDs (string key-value pairs)",
-                              "additionalProperties": {
-                                "type": "string"
-                              },
-                              "example": {"payrollId": "123", "internalDriverId": "250020"}
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "externalIds" : {
+                                    "type": "object",
+                                    "description": "Dictionary of external IDs (string key-value pairs)",
+                                    "additionalProperties": {
+                                        "type": "string"
+                                    },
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4014,40 +4014,97 @@
                     "type": "string",
                     "description": "Reason that this driver is exempt from the Electronic Logging Mandate (see eldExempt)."
                 },
-                "eldBigDayExemptionEnabled": {
+                "usBigDayExemptionEnabled": {
```
AChelikani: We will need to duplicate these fields and leave the existing ones to not break current implementations.

alberthu16: I'll leave the existing ones there in our actual API implementation, but delete them here in the documentation.

swagger.json
```diff
@@ -4042,6 +4042,69 @@
...
+                    "type": "integer",
+                    "description": "Carrier’s US DOT number (max: 999999999), if different from the organization default.",
+                    "example": "999999999"
+                },
+                "tags": {
+                    "type": "array",
+                    "description": "An array of tags that the driver belongs to.",
+                    "items": {
+                        "type": "string",
+                        "example": "ExampleTag"
```
AChelikani: It might be better to put a better example tag here maybe like a tag that ArcBest currently uses?

swagger.json
```diff
@@ -4187,28 +4187,24 @@
...
-                                "type": "integer",
-                                "format": "int64",
-                                "description": "ID of the driver.",
-                                "example": 444
-                            },
-                            "name": {
-                                "type": "string",
-                                "description": "Name of the driver.",
-                                "example": "Fred Jacobs"
+                        "allOf": [
```
AChelikani: This is just something to think about, not really sure if it would really cause issues.
```
{
  "drivers": [
    {
      "id": 444,
      "name": "Fred Jacobs"
    }
  ]
}
```
Since this increases the response payload significantly and our API doesn't have pagination it may cause orgs with a lot of drivers get timeouts when they hit this endpoint.

alberthu16: Good consideration -- I chatted with @silverspace and the short answer is that we don't need to be concerned about returning the fat driver object here.

We currently use https://datastudio.google.com/reporting/1miK3rbBvE_kvGrck08gquBBxZ3xy-buK/page/4pKV to track metrics.

And I think at some point in the future we'll want to support toggling fat vs skinny response for getting these collection resources via query param.

PR #93, opened 2018-12-06
Add driveMs, onDutyMs, pendingDriveMs, and pendingOnDutyMs to hos_logs_summary

faizandfurious:


swagger.json
```diff
@@ -4430,7 +4430,31 @@
                     "format": "int64",
                     "description": "The amount of driving time in violation in this cycle (in ms).",
                     "example": 50400000
-                  }
+                  },
+                  "onDutyMs": {
```
silverspace: Nit: this looks like a unix timestamp until you read the description. Can we call this `onDutyDurationMs`? 

Also, it's not clear whether this is the current shift, cycle, etc, unless you read the description. Maybe `currentCycle { onDutyDurationMs }`?

swagger.json
```diff
@@ -4430,7 +4430,31 @@
...
+                    "type": "number",
+                    "format": "int64",
+                    "description": "The amount of on duty time in this cycle (in ms).",
```
silverspace: nit: s/ms/milliseconds

swagger.json
```diff
@@ -4430,7 +4430,31 @@
...
+                  "driveMs": {
+                    "type": "number",
+                    "format": "int64",
+                    "description": "The amount of driving time in this cycle (in ms).",
+                    "example": 34200000
+                  },
+                  "pendingOnDutyMs": {
+                    "type": "number",
+                    "format": "int64",
+                    "description": "The amount of on duty time in this cycle for pending logs (in ms).",
```
silverspace: What is a pending log? Does the above OnDutyDurationMs value include pending logs?

faizandfurious: I can rephrase this. `pendingOnDutyMs` represents the amount of on duty time in this cycle, including any pending logs.

swagger.json
faizandfurious: The naming convention here is consistent with the naming convention in `hos_daily_logs`

faizandfurious: But I see that there's a few different time periods represented in this endpoint. @dsharrison any opinion on naming?

dsharrison: Good question. I don't have a strong preference, but think we should favor consistency within our API.

faizandfurious:
Opening a new PR for this.

PR #94, opened 2018-12-06
Add documentation for GET /fleet/drivers

alberthu16:


swagger.json
```diff
@@ -725,6 +725,48 @@
...
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "name": "group_id",
+                        "required": false,
+                        "in": "query",
+                        "type": "integer",
+                        "format": "int64",
+                        "description": "Optional Group ID to query for.",
```
alberthu16: In the ApiHandler, we'll either use this query param passed in from the request, or we'll default to the one group that's usually associated with all orgs.

swagger.json
```diff
@@ -680,37 +680,33 @@
...
+                        "example": 10566
                     }
                 ],
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/drivers",
+                "description": "Get all the drivers for the specified group.",
```
silverspace: Let's remove the `group` nomenclature here. Most orgs only have one group and don't know what a group means. Also, can we use the verb `List` here instead of `Get`? 

Maybe `List all the drivers in the fleet.`?

swagger.json
```diff
@@ -680,37 +680,33 @@
...
-                            "properties": {
-                                "groupId": {
-                                    "type": "integer",
-                                    "format": "int64",
-                                    "description": "Group ID to query.",
-                                    "example": 101
-                                }
-                            }
-                        }
+                        "name": "group_id",
```
silverspace: This can be 
```
                    { "$ref": "#/parameters/groupIdParam" }
```

swagger.json
```diff
@@ -680,37 +680,33 @@
...
+                        "description": "Optional Group ID to query for.",
+                        "example": 10566
                     }
                 ],
+                "tags": [
+                    "Fleet", "Default"
+                ],
+                "summary": "/fleet/drivers",
+                "description": "Get all the drivers for the specified group.",
+                "operationId": "getFleetDrivers",
```
silverspace: Nit: can we also change this to `listFleetDrivers`?

swagger.json
alberthu16: sounds good

PR #100, opened 2018-12-24
Remove default category from tags

katielee-samsara:
Remove default as a category from tags, to simplify samsara.com/api documentation categories

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -2120,7 +2120,7 @@
         "/tags": {
             "get": {
                 "tags": [
-                    "Fleet", "Industrial"
+                    "Tags"
                 ],
                 "summary": "/tags",
                 "description": "Fetch all of the tags for a group.",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -2321,7 +2308,7 @@
             },
             "delete": {
                 "tags": [
-                    "Default"
+                    "Fleet", "Industrial"
```
silverspace: Did you intend to add both Fleet and Industrial to these tags endpoints? Maybe we should create a separate Tags tag?

katielee-samsara: I assumed it would be easiest for Industrial customers to easily go to one spot to find all the APIs that are relevant to industrial, assuming everything else is Fleet-only. Per second look, it seems that Sensors are also related to Industrial and not in the Industrial section, so I'll similarly create a separate Tags section.

PR #101, opened 2018-12-28
add addresses to tag GET endpoints

EmmaOnThursday:
Add address object to tag single GET and tag LIST responses. 

View docs here: https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/d1bf43ea37a13a68ae8f607ddf6895bdec41762b/swagger.json#operation/GetOrganizationContact

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -4362,6 +4362,37 @@
                 }
             ]
         },
+        "TaggedAddressBase": {
+            "type": "object",
+            "required": [
+                "id"
+            ],
+            "properties": {
+                "id": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "The ID of the address being tagged.",
+                    "example": 345
+                }
+            }
+        },
+        "TaggedAddress": {
+            "allOf": [
+                {
+                    "$ref": "#/definitions/TaggedAddressBase"
+                },
+                {
+                    "type": "object",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4362,6 +4362,37 @@
...
+        },
+        "TaggedAddress": {
+            "allOf": [
+                {
+                    "$ref": "#/definitions/TaggedAddressBase"
+                },
+                {
+                    "type": "object",
+                    "properties": {
+                        "name": {
```
silverspace: Should `name` be required?

EmmaOnThursday: we don't require `name` in the other `Tagged<object>` definitions.

silverspace: Ok, we can fix them all in a separate PR.

swagger.json
```diff
@@ -4362,6 +4362,37 @@
                 }
             ]
         },
+        "TaggedAddressBase": {
```
silverspace: Is `TaggedAddressBase` only for `TagCreate` purposes? (Asking because this is missing all of the other address related fields, like geofence, etc).

EmmaOnThursday: It's actually only for the `GET` endpoint, single & list.

PR #104, opened 2019-01-02
swagger: add fleet/vehicle endpoint

seward:
Add the fleet/vehicle endpoint with GET and PATCH methods.  Also adds a FleetVehicleResponse definition, since the "vehicle" returned by this endpoint differs from the existing Vehicle definition.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1641,7 +1641,94 @@
                 }
             }
         },
-      "/fleet/vehicles/locations": {
+        "/fleet/vehicles/{vehicle_id_or_external_id}": {
+            "get": {
+                "tags": [
+                    "Fleet"
+                ],
+                "summary": "/fleet/vehicles/{vehicle_id_or_external_id}",
+                "description": "Gets a specific vehicle.",
+                "operationId": "getFleetVehicle",
+                "consumes": [
+                    "application/json"
+                ],
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "$ref": "#/parameters/vehicleIdOrExternalId"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "The specified vehicle.",
+                        "schema": {
+                            "$ref": "#/definitions/FleetVehicleResponse"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": ["Fleet"],
+                "summary": "/fleet/vehicles/{vehicle_id_or_external_id}",
+                "description": "Updates the specified vehicle using JSON merge patch format. See IETF RFC 7396: https://tools.ietf.org/html/rfc7396.",
+                "operationId": "patchFleetVehicle",
+                "consumes": [
+                    "application/json",
+                    "application/merge-patch+json"
+                ],
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "$ref": "#/parameters/vehicleIdOrExternalId"
+                    },
+                    {
+                        "in": "body",
+                        "name": "data",
+                        "required": true,
+                        "schema": {
+                            "type": "object",
+                            "properties":
+                            {
+                                "name": {
+                                    "description": "Name",
+                                    "type": "string"
+                                },
+                                "harsh_accel_setting": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -1641,7 +1641,87 @@
                 }
             }
         },
-      "/fleet/vehicles/locations": {
+        "/fleet/vehicles/{vehicle_id_or_external_id}": {
+            "get": {
+                "tags": [
+                    "Fleet"
+                ],
+                "summary": "/fleet/vehicles",
```
silverspace: Wrong summary string here.

swagger.json
```diff
@@ -1641,7 +1641,87 @@
...
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": ["Fleet"],
+                "summary": "/fleet/vehicles",
```
silverspace: Here as well.

swagger.json
```diff
@@ -1641,7 +1641,87 @@
...
+                "description": "Updates the specified vehicle",
```
silverspace: Could we also elaborate on the patch semantics? We should describe that users should use https://tools.ietf.org/html/rfc7396 

This includes the appropriate `Content-Type: application/merge-patch+json` and JSON Patch body.

swagger.json
```diff
@@ -1641,7 +1641,87 @@
...
+                "parameters": [
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "$ref": "#/parameters/vehicleIdOrExternalId"
+                    },
+                    {
+                        "in": "body",
+                        "name": "data",
```
silverspace: Should we use the name `data` here, or `patchedVehicle`?

swagger.json
```diff
@@ -3007,6 +3087,64 @@
...
+                    "description": "ID of the vehicle.",
+                    "format": "int64",
+                    "example": 112
+                },
+                "name": {
+                    "type": "string",
+                    "description": "Name of the vehicle.",
+                    "example": "Truck A7"
+                },
+                "vehicleInfo": {
```
silverspace: Nit: Does this endpoint already exist and return `vehicleInfo`? This seems like an unnecessary envelope to me.

swagger.json
```diff
@@ -6418,6 +6556,13 @@
             "in": "query",
             "type": "integer",
             "format": "int64"
+        },
+        "vehicleIdOrExternalId": {
+            "name": "vehicle_id_or_external_id",
+            "description": "ID of the vehicle.  This must be either the numeric ID or the external ID of the vehicle.",
```
silverspace: Could we elaborate that the numerical id is the Samsara generated identifier for the vehicle, and external ids are customer specified key=value pairs that can be defined for a vehicle using the PATCH endpoint?

swagger.json
seward: Yeah, the endpoint returns a list of devices like 
```graphql
type Device struct {
	Id          int64
	Name        string
	VehicleInfo struct {
		Vin   string
		Make  string
		Model string
		Year  int64
	}
	HarshAccelSetting int64
	ExternalIds       []struct {
		Key   string
		Value string
	}
}
```

swagger.json
seward: Would it be correct to say that this endpoint consumes like
```                
"consumes": [
    "application/json",
    "application/merge-patch+json"
 ],
```
for PATCH, or only `application/merge-patch+json`?

swagger.json
seward: It only seems to matter for how you use the autogenerated client.  I was wavering between data or body, so that querying the endpoint is something like 
```
apiClient.Fleet.PatchFleetVehicle(fleet.NewPatchFleetVehicleParams().
    WithAccessToken(...).
    WithVehicleIDOrExternalID(...).
    WithData(...<body params>...))
```
but as to what it should be it really depends on how we want people to use the client.  I prefer something more regular for adding to the body, so `data` or `body`, but if we have an existing convention then it makes sense to change it to `patchedVehicle`.

silverspace: SGTM!

swagger.json
```diff
@@ -1641,7 +1641,94 @@
...
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+                "tags": ["Fleet"],
+                "summary": "/fleet/vehicles/{vehicle_id_or_external_id}",
+                "description": "Updates the specified vehicle using JSON merge patch format.",
```
silverspace: Could we also link to the RFC page here?

PR #105, opened 2019-01-03
swagger: add docs for vehicle/stats endpoint

sjain28:
Adds docs for a `vehicle/id/stats` endpoint that returns engine state and digio (digital aux) objectStats within a timerange (no greater than 1 hour). 

https://paper.dropbox.com/doc/Mini-Dev-Spec-Pike-Electric-API-Updates--AU4cCkqxlRH7SrVwwUjOiCMVAg-wzPIvSmJf8MVofdtJXml6

Dev work on this endpoint is not complete, but Pike wanted to see an API spec.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1821,6 +1821,84 @@
             }
           }
         },
+        "/fleet/vehicles/stats": {
+              "get": {
+                  "tags": ["Fleet"],
+                  "summary": "/fleet/vehicles/stats",
+                  "description": "Fetch engine state and aux input data for all vehicles in the group between a start/end time. Data returned may be affected by device connectivity and processing time.",
+                  "operationId": "GetVehicleStats",
+                  "parameters": [
+                    {
+                      "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                      "name": "startMs",
+                      "description": "Time in Unix epoch milliseconds for the start of the query.",
+                      "required": true,
+                      "in": "query",
+                      "type": "integer"
+                    },
+                    {
+                      "name": "endMs",
+                      "description": "Time in Unix epoch milliseconds for the end of the query.",
+                      "required": true,
+                      "in": "query",
+                      "type": "integer"
+                    },
+                    { 
+                      "name": "series",
+                      "description": "Comma-separated list of stat types. Options are engineState, auxInput1, and auxInput2. If this parameter is excluded, all 3 stat types will be returned. Example: series=engineState,auxInput2",
+                      "type": "string",
+                      "in": "query",
+                      "enum": ["engineState", "auxInput1", "auxInput2"]
+                    }
+                  ],
+                  "responses": {
+                    "200": {
+                      "description": "Returns engine state and/or aux input data for all vehicles in the group between a start/end time.",
+                      "schema": {
+                        "type": "object",
+                        "required": ["vehicleStats"],
+                        "properties": {
+                            "vehicleStats": {
+                                "type": "array",
+                                "items": {
+                                    "type": "object",
+                                    "required": ["vehicleId"],
+                                    "properties": {
+                                        "vehicleId": {
+                                            "type": "integer",
+                                            "description": "ID of the vehicle.",
+                                            "format": "int64",
+                                            "example": 112
+                                        },
+                                        "engineState": {
+                                            "type": "array",
+                                            "items": {
+                                                "$ref": "#/definitions/EngineState"
+                                            }
+                                        },
+                                        "auxInput1": {
+                                            "$ref": "#/definitions/AuxInputSeries"
+                                        },
+                                        "auxInput2": {
+                                            "$ref": "#/definitions/AuxInputSeries"
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                      }
+                    },
+                    "default": {
+                      "description": "Unexpected error.",
+                      "schema": {
+                        "$ref": "#/definitions/ErrorResponse"
+                      }
+                    }
+                  }
+              }
+        },
         "/fleet/drivers/documents": {
             "parameters": [
                 {
@@ -3102,6 +3180,59 @@
                     "format": "double",
                     "description": "Heading in degrees.",
                     "example": 246.42
+                }
+            }
+        },
+        "AuxInput": {
+            "type": "object",
+            "description": "Digital value of an aux input.",
+            "required": ["timeMs", "value"],
+            "properties": {
+                "timeMs": {
+                    "type": "number",
+                    "description": "Timestamp in Unix epoch milliseconds.",
+                    "format": "int64",
+                    "example": "1546542978484"
+                },
+                "value": {
+                    "type": "boolean",
+                    "description": "Boolean representing the digital value of the aux input.",
+                    "example": true
+                }
+            }
+        },
+        "AuxInputSeries": {
+            "type": "object",
+            "description": "A list of aux input values over a timerange.",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "The name of the aux input.",
+                    "example": "Boom",
+                    "enum": ["Emergency Lights", "Emergency Alarm", "Stop Paddle", "Power Take-Off", "Plow", "Sweeper", "Salter", "Boom"]
+                },
+                "values": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/AuxInput"
+                    }
+                }
+            }
+        },
+        "EngineState": {
+            "type": "object",
+            "description": "The state of the vehicle over time. State can be Running, Off, or Idle.",
+            "required": ["timeMs", "value"],
+            "properties": {
+                "timeMs": {
+                    "type": "number",
+                    "description": "Timestamp in Unix epoch milliseconds.",
+                    "format": "int64",
+                    "example": "1546542978484"
+                },
+                "value": {
+                    "type": "string",
+                    "enum": ["Running", "Off", "Idle"]
                 }
             }
         },


swagger.json
```diff
@@ -1821,6 +1821,57 @@
             }
           }
         },
+        "/fleet/vehicles/{vehicle_id}/stats": {
```
oliversong: out of curiosity, is there an interface or generator to create this json? or do we have to hand write it?

katielee-samsara: Nit: the current format is {vehicle_id:[0:9]+} for the other endpoints

swagger.json
```diff
@@ -1821,6 +1821,57 @@
...
+                  "required": true,
+                  "in": "query",
+                  "type": "integer",
+                  "format": "int64"
+                }
+              ],
+              "get": {
+                  "tags": ["Fleet"],
+                  "summary": "/fleet/vehicles/{vehicle_id:[0-9]+}/stats",
+                  "description": "Fetch engine state and digital aux data for a given vehicle between a start/end time. The maximum query duration is one hour.",
```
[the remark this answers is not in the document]
katielee-samsara: Left comment within [mini dev spec](https://paper.dropbox.com/doc/Mini-Dev-Spec-Pike-Electric-API-Updates--AU8GqZ4wvVh5~ma2YfbYZ5u5Ag-wzPIvSmJf8MVofdtJXml6). I'm curious to know if this approach is scalable, and if it would be possible to instead have the developer pass in parameters to request certain object stats.

swagger.json
```diff
@@ -3105,6 +3156,66 @@
                 }
             }
         },
+        "VehicleStats": {
+            "type": "object",
+            "description": "Engine state and digital aux data for a vehicle.",
+            "properties": {
+                "engineState": {
+                    "$ref": "#/definitions/ObjectStatIntSeries"
```
katielee-samsara: Left comment in mini dev spec. Could you add more information on what is the expected response array size? Are object stats created every x seconds, or do we only show object stats for a state change for these boolean values?

swagger.json
sjain28: We have to hand write it at the moment.

swagger.json
```diff
@@ -1821,6 +1821,57 @@
...
+                  "description": "Fetch engine state and digital aux data for a given vehicle between a start/end time. The maximum query duration is one hour.",
+                  "operationId": "GetVehicleStats",
+                  "responses": {
+                    "200": {
+                      "description": "Returns engine state and digital aux data for a given vehicle between a start/end time, with a maximum query duration of one hour.",
+                      "schema": {
+                        "$ref": "#/definitions/VehicleStats"
+                      }
+                    },
+                    "default": {
```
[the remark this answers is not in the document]
sjain28: In this case, `default` describes the schema of the API response for any code other than 200. In the `responses` object we can enumerate response types for different status codes, and have a  `default` option to handle all the remaining status codes.

swagger.json
```diff
@@ -3105,6 +3156,66 @@
                 }
             }
         },
+        "VehicleStats": {
```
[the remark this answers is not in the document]
sjain28: No formal order, but yeah. I can move the object stats stuff to the bottom so that they are grouped together.

swagger.json
sjain28: I have `{vehicle_id:[0:9]+} ` in the `summary` for the `stats` endpoint that I'm adding. It doesn't look like we add the regex in the list of routes here, see L1765.

swagger.json
sjain28: For `engineState`, we only show object stats for a state change.

swagger.json
```diff
@@ -3105,6 +3156,66 @@
...
+                },
+                "value": {
+                    "type": "integer",
+                    "description": "Integer value.",
+                    "format": "int64",
+                    "example": "1030"
+                }
+            }       
+        },
+        "ObjectStatIntSeries": {
```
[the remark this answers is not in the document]
sjain28: Weird, looks fine here. 

screen shot 2019-01-04 at 10 38 20 am

swagger.json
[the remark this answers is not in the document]
sjain28: Actually I'd prefer to keep them where they are, grouped with `VehicleLocation` since that response type also contains objectstats.

sjain28:
@katielee-samsara I tweaked some of the language and addressed your comments above and on the dev spec.

swagger.json
katielee-samsara: Is this the same for digital aux data?

sjain28: yes

sjain28:
@silverspace made the following changes after our conversation: 
* Changed route to `/vehicles/stats` so that we return stats for all vehicles
* Added `filter` argument
* Changed "digital aux" to "digital input" to match cloud dashboard language

swagger.json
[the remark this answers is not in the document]
sjain28: Added a `filter` argument where caller can request particular stats.

swagger.json
```diff
@@ -1821,6 +1821,69 @@
...
+                  "name": "startMs",
+                  "description": "Time in Unix epoch milliseconds for the start of the query (cannot exceed 1 hour)",
+                  "required": true,
+                  "in": "query",
+                  "type": "integer",
+                  "format": "int64"
+                },
+                {
+                  "name": "endMs",
+                  "description": "Time in Unix epoch milliseconds for the end of the query (cannot exceed 1 hour)",
```
silverspace: Same as above. Would we always return the most recent timeseries value if we passed graphql (nowMs, nowMs) as the (startMs, endMs) values? If so, that would be an ideal default.

swagger.json
```diff
@@ -1821,6 +1821,69 @@
...
         },
+        "/fleet/vehicles/stats": {
+            "parameters": [
+                {
+                  "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                  "name": "startMs",
+                  "description": "Time in Unix epoch milliseconds for the start of the query (cannot exceed 1 hour)",
+                  "required": true,
```
silverspace: Do we want to make startMs and endMs required params? For instance, we could default endMs to `now`.

swagger.json
```diff
@@ -1821,6 +1821,69 @@
...
+                {
+                  "name": "endMs",
+                  "description": "Time in Unix epoch milliseconds for the end of the query (cannot exceed 1 hour)",
+                  "required": true,
+                  "in": "query",
+                  "type": "integer",
+                  "format": "int64"
+                },
+                { 
+                  "name": "filter",
```
silverspace: Sorry for changing the naming convention here -- could we rename this to `series`?

swagger.json
```diff
@@ -1821,6 +1821,69 @@
...
+                  "description": "Comma-separated list of stat types. Options are engineState, auxInput1, and auxInput2.",
```
silverspace: Should we just have a single `aux-inputs` type that returns both `auxInput1` and `auxInput2`?

silverspace: I think this should be a required query parameter, since we want to avoid returning all timeseries data if this API gets larger.

swagger.json
```diff
@@ -1821,6 +1821,69 @@
             }
           }
         },
+        "/fleet/vehicles/stats": {
+            "parameters": [
+                {
+                  "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                  "name": "startMs",
```
silverspace: These parameter should all be in the "get" method section.

swagger.json
```diff
@@ -1821,6 +1821,69 @@
...
+              "get": {
+                  "tags": ["Fleet"],
+                  "summary": "/fleet/vehicles/stats",
+                  "description": "Fetch engine state and aux input data for all vehicles in the group between a start/end time. The maximum query duration is one hour. Data returned may be affected by device connectivity and processing time.",
+                  "operationId": "GetVehicleStats",
+                  "responses": {
+                    "200": {
+                      "description": "Returns engine state and aux input data for all vehicles in the group between a start/end time, with a maximum query duration of one hour.",
+                      "schema": {
+                        "type": "array",
```
silverspace: Could we return a dictionary at the top level rather than an array? That way the result type is extensible if we want to add additional metadata in the future.

swagger.json
```diff
@@ -3105,6 +3168,76 @@
...
+            "properties": {
+                "timeMs": {
+                    "type": "integer",
+                    "description": "Timestamp in Unix epoch milliseconds.",
+                    "format": "int64",
+                    "example": "1546542978484"
+                },
+                "value": {
+                    "type": "integer",
+                    "description": "Integer value.",
```
silverspace: Nit: This description is not helpful, could we add some more description about what this value represents for different aux types?

swagger.json
```diff
@@ -3105,6 +3168,76 @@
...
+        },
+        "ObjectStatIntSeries": {
+            "type": "array",
+            "items": {
+                "$ref": "#/definitions/ObjectStatInt"
+            }
+        },
+        "ObjectStatEngineState": {
+            "type": "object",
+            "description": "The state of the vehicle over time. 0 = Off, 1 = Running, 2 = Idle.",
```
silverspace: Could this be a swagger enum? https://swagger.io/docs/specification/2-0/enums/

Could we also use the human readable name rather than the int value?

swagger.json
```diff
@@ -3105,6 +3168,76 @@
...
+                "engineState": {
+                    "$ref": "#/definitions/ObjectStatEngineState"                    
+                },
+                "auxInput1": {
+                    "type": "object",
+                    "required": ["name", "values"],
+                    "properties": {
+                        "name": {
+                            "type": "string",
+                            "example": "Boom"
```
silverspace: Could we enumerate the DIGIO input types here? https://swagger.io/docs/specification/2-0/enums/

swagger.json
```diff
@@ -1821,6 +1821,69 @@
...
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "vehicleId": {
+                                    "type": "integer",
+                                    "description": "ID of a vehicle",
+                                    "format": "int64",
+                                    "example": "33"
+                                },
+                                "stats": {
```
silverspace: Is this stats wrapper required, or should we directly inline the relevant series here?

swagger.json
sjain28: Both are required in `/fleet/vehicles/locations`, and I'd prefer to have `/stats` match that endpoint.

swagger.json
sjain28: Renamed

swagger.json
sjain28: Moved the parameters.

swagger.json
sjain28: Changed the return type

swagger.json
sjain28: Yep, changed to be an enum of human-readable values.

swagger.json
sjain28: Inlined the timeseries objects.

swagger.json
```diff
@@ -1821,6 +1821,80 @@
...
+                    },
+                    {
+                      "name": "endMs",
+                      "description": "Time in Unix epoch milliseconds for the end of the query.",
+                      "required": true,
+                      "in": "query",
+                      "type": "integer"
+                    },
+                    { 
+                      "name": "series",
```
oliversong: should this be `filter` ?

sjain28: Renamed it to `series` as per Greg's comment earlier.

oliversong: or, should the example be `series=engineState,auxInput2`?

sjain28: changed the example

PR #107, opened 2019-01-08
api docs: add docs for vehicle patch body

AChelikani:
Update vehicle PATCH docs for arcbest commit to include patch body payload.


seward:
LGTM

PR #111, opened 2019-01-14
swagger.json: Add tags to driver create and response

Stonelinks:


swagger.json
```diff
@@ -3828,6 +3828,9 @@
                             "type": "string",
                             "description": "Driver's password for the driver app.",
                             "example": "mypassword"
+                        },
+                        "tagIds": {
+                            "$ref": "#/definitions/TagIds"
```
silverspace: Is it possible for drivers to have tag specific roles?

Stonelinks: I don't think so... the only thing I know about are tag admins who can set a tag that basically lock a user within that tag. Doesn't change their role or anything.

PR #112, opened 2019-01-14
swagger.json: update DriversResponse to use Driver objects

Stonelinks:
now that the active and inactive fleet driver list endpoints are returning the same response, the `DriversResponse` can be made to use the `Driver` object


Stonelinks:
hold off on merging this until the corresponding backend PR is merged

PR #113, opened 2019-01-15
[rfc] update tagRoles to use ids/uuids instead of the role names

ehzhang:
Docs preview: https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/ehzhang/arcbest-preview-user-adjustments/swagger.json#operation/listUserRoles

For writes (POST, PATCH, PUT), using the name of a tag as a unique identifier is untenable. We should instead change this to use the tagId instead of tagName.

For role, we can actually keep using the role name (since those are effectively an enum, there's a set of specific standard roles) but we may want to introduce a custom role uuid.

From:
```
{
  "name": "Jane Doe",
  "email": "user@company.com",
  "authType": "basic",
  "organizationRole": "Read-Only Admin",
  "tagRoles": [
    {
      "tagName": "North Carolina Distribution",
      "role": "Read-Only Admin"
    }
  ]
}
```

We've decided to settle on a payload of the shape:
```
{
  "name": "Jane Doe",
  "email": "user@company.com",
  "authType": "default",
  "organizationRoleId": "8a9371af-82d1-4158-bf91-4ecc8d3a114c",
  "tagRoles": [
    {
      "tagId": "123",
      "roleId": "8a9371af-82d1-4158-bf91-4ecc8d3a114c"
    }
  ]
}
```

This has the notable change in user behavior where api clients will need to fetch for the tags using our `tags` endpoint in order to pass in the correct id, as opposed to copying down the name of the tag (either by fetching using our `tags` endpoint or by copying it down manually from the dashboard). Using an id instead of a tag name is guaranteed to be unique, as tag names are easily mutable.

We will also need to extend a new `roles/` endpoint that exposes role uuids (ids?) that clients will need to hit in order for them to get the appropriate uuid to attach. This means that api clients won't have to keep in sync manually the name of a custom role, since role names are likewise mutable.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -47,6 +47,36 @@
         }
     ],
     "paths": {
+        "/user_roles": {
+            "get": {
+                "tags": [
+                    "Users"
+                ],
+                "summary": "/user_roles",
+                "description": "Get all roles in the organization.",
+                "operationId": "listUserRoles",
+                "parameters": [
+                    { "$ref": "#/parameters/accessTokenParam" }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "List of user roles.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/UserRole"
+                            }
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
         "/users": {
             "get": {
                 "tags": [
@@ -78,26 +108,30 @@
             },
             "post": {
                 "tags": [
-                    "Default", "Users"
+                    "Users"
                 ],
                 "summary": "/users",
                 "description": "Add a user to the organization.",
-                "operationId": "create_user",
+                "operationId": "createUser",
                 "parameters": [
-                    { "$ref": "#/parameters/accessTokenParam" },
-                    { "$ref": "#/parameters/createUserParam" }
+                    {
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "$ref": "#/parameters/createUserParam"
+                    }
                 ],
                 "responses": {
                     "200": {
                         "description": "User successfully created.",
                         "schema": {
-                            "$ref": "#/definitions/User"
+                        "$ref": "#/definitions/User"
                         }
                     },
                     "default": {
                         "description": "Unexpected error.",
                         "schema": {
-                            "$ref": "#/definitions/ErrorResponse"
+                        "$ref": "#/definitions/ErrorResponse"
                         }
                     }
                 }
@@ -169,16 +203,20 @@
             },
             "patch": {
                 "tags": [
-                    "Default", "Users"
+                  "Users"
                 ],
-                "summary": "/users/{user_id:[0-9]+}",
+                "summary": "/users/{userId:[0-9]+}",
                 "description": "Update some fields on a user. Only fields to be changed need to be supplied. Note that if you change an object or collection, you must supply the complete collection. For example, to add a tag role to a user, you must specify all tag roles that the user should have.",
-                "operationId": "update_user",
+                "operationId": "patchUser",
                 "parameters": [
-                    { "$ref": "#/parameters/accessTokenParam" },
-                    { "$ref": "#/parameters/updateUserParam" },
                     {
-                        "name": "user_id",
+                        "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                        "$ref": "#/parameters/updateUserParam"
+                    },
+                    {
+                        "name": "userId",
                         "in": "path",
                         "required": true,
                         "description": "ID of the user.",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -6855,15 +6855,26 @@
...
-                                    "role" : {
+                                    "roleId" : {
+                                        "type": "number",
+                                        "description": "The id of the standard role type",
+                                        "example": "123f"
+                                    },
+                                    "roleName" : {
                                         "type": "string",
-                                        "description": "The name of the role the user is assigned to for this tag. This must match one of the preset roles or the name of a custom role in the organization. This value is case insensitive.",
+                                        "description": "The name of the standard role type (case insensitive)",
```
silverspace: Per our in-person discussion, could we define the enum of possible values for `roleName`, and drop `roleId`?

ehzhang: yup, will update this PR accordingly (i was also pretty lax on description until we talked :P)

ehzhang: added enum roles

swagger.json
```diff
@@ -6855,15 +6855,30 @@
...
+                                            "Maintenance"
+                                        ]
                                     },
-                                    "role" : {
+                                    "customRoleUuid" : {
                                         "type": "string",
-                                        "description": "The name of the role the user is assigned to for this tag. This must match one of the preset roles or the name of a custom role in the organization. This value is case insensitive.",
-                                        "example": "Read-Only Admin"
+                                        "description": "The uuid of the custom role the user is assigned to for this tag. Either this field or roleId must be provided.",
+                                        "example": "8a9371af-82d1-4158-bf91-4ecc8d3a114c"
```
dsharrison: Are we tracking the addition of custom role `/GET` in another ticket?

dsharrison: Should we refer to this as just `id`? Do our customers care about the difference between types of identifiers?

swagger.json
```diff
@@ -6855,15 +6855,30 @@
...
+                                        "description": "The id of the tag to grant this user access to.",
+                                        "example": "123"
+                                    },
+                                    "roleName" : {
                                         "type": "string",
-                                        "description": "The name of the tag to grant this user access to.",
-                                        "example": "North Carolina Distribution"
+                                        "description": "The name of the standard role type (case insensitive). Either this field or customRoleUuid must be provided.",
+                                        "example": "Read-Only Admin",
+                                        "enum": [
```
dsharrison: I like that this approach provides easy access to built-in roles. However, if a user already has to make a `/GET` call for any custom role values, should we simply include standard roles in that list and accept just an `id` for both standard and custom here?

You may have already discussed with with @silverspace so feel free to just respond with "We already discussed and landed on this approach" 😄

ehzhang: > Do we have a reserved name validation on custom roles, so customers cannot select standard role names like "Full Admin"?

customers cannot change or create roles that match the names of existing standard roles

ehzhang: with generated sdk type things and for our api documentation in general, do we want to allow for this kind of "ids can be either strings or numbers" behavior?

ehzhang: I might offer that we _could_ hardcode uuids for our existing role types (rather than use number ids) so that the api is consistent

ehzhang: Assuming we create a `roles/` endpoint, might we change it to instead:
```
{
  "name": "Jane Doe",
  "email": "user@company.com",
  "authType": "basic",
  "organizationRole": "Read-Only Admin",
  "tagRoles": [
    {
      "tagId": "123",
      "roleId": "1"
    },
    {
      "tagId": "123",
      "roleId": "aaaa-aaaa-aaa-aaa"
    },
  ]
}
```

silverspace: Yes I am fine with treating the id as a string. I like this over creating a `uuid` for standard roles just for API.

ehzhang: sounds good. I'll change this to use only the `roleId` instead of using `roleName`

ehzhang: adjusted

swagger.json
ehzhang: we've changed this to now be just the id (number or uuid format)

ehzhang: added a get endpoint for `/user_roles`

ehzhang:
@silverspace @dsharrison adjusted the proposal - preview here:
https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/ehzhang/arcbest-preview-user-adjustments/swagger.json#operation/listUserRoles

dsharrison:
@ehzhang I think we should update `organizationRole` to `organizationRoleId` on the `PATCH` as well.

ehzhang:
@silverspace would it be too weird to have the return type not quite match up?

ehzhang:
We've decided to settle on a payload of the shape:
```
{
  "name": "Jane Doe",
  "email": "user@company.com",
  "authType": "default",
  "organizationRoleId": "8a9371af-82d1-4158-bf91-4ecc8d3a114c",
  "tagRoles": [
    {
      "tagId": "123",
      "roleId": "8a9371af-82d1-4158-bf91-4ecc8d3a114c"
    }
  ]
}
```

ehzhang:
@dsharrison can you take another pass at this arcbest preview for /users CRUD?

dsharrison:
This still LGTM as long as we are returning standard roles from `GET /user_roles` to be passed to the `organizationRoleId` parameter in the patch.

ehzhang:
@dsharrison yup, we return all standard/default roles (not including samsara roles, of course!)

PR #118, opened 2019-01-23
swagger hos_certified_daily_log: Adding shipping id & trailer number

sskhandek:
Adding docs for `/fleet/drivers/{driver_id}/hos_certified_daily_log/{time_ms}`

More details and discussion of edge cases here: **https://paper.dropbox.com/doc/Arc-Best-API-Edit-Form-And-Manner--AWKkjfFAdBwu1bMWjxhY2Lq7Ag-dhVr6zVFqHZhUAmq2GBXD**

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1048,7 +1048,7 @@
           }
         },
         "/fleet/drivers/{driver_id}/hos_daily_logs": {
-          "post": {
+          "get": {
             "tags": [
               "Fleet"
             ],
--- a/swagger.json
+++ b/swagger.json
@@ -1121,6 +1121,122 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/hos_certified_daily_logs/{time_ms}": {
+            "get": {
+                "tags": [
+                    "Fleet", "Drivers"
+                ],
+                "summary": "/fleet/drivers/{driver_id:[0-9]+}/hos_certified_daily_logs/{time_ms:[0-9]+}",
+                "description": "Fetch a certified daily log for driver by driver_id and timestamp.",
+                "operationId": "get_hos_certified_daily_logs",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "parameters": [
+                    {
+                      "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                      "description": "Identifier of the driver whose certified daily log is being updated",
+                      "in": "path",
+                      "name": "driver_id",
+                      "required": true,
+                      "type": "integer"
+                    },
+                    {
+                        "description": "A specified timestamp within the bounds of a driver's certified daily log, specified in milliseconds UNIX time.",
+                        "in": "path",
+                        "name": "time_ms",
+                        "required": true,
+                        "type": "integer"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Returns a certified daily log for the driver, if one exists",
+                        "schema": {
+                            "$ref": "#/definitions/HosCertifiedDailyLogResponse"
+                        }
+                    },
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+              "summary": "/fleet/drivers/{driver_id:[0-9]+}/hos_certified_daily_logs/{time_ms:[0-9]+}",
+              "description": "Modify details, such as trailer number or shipping ID, for a driver's certified daily log",
+              "operationId": "patch_hos_certified_daily_log",
+              "tags": [
+                "Fleet", "Drivers"
+              ],
+              "produces": [
+                "application/json"
+              ],
+              "consumes": [
+                "application/json"
+              ],
+              "parameters": [
+                {
+                  "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                  "description": "ID (integer) of the driver with certified daily log.",
+                  "in": "path",
+                  "name": "driver_id",
+                  "required": true,
+                  "type": "integer"
+                },
+                {
+                    "description": "A specified timestamp within the bounds of a driver's certified daily log, specified in milliseconds UNIX time.",
+                    "in": "path",
+                    "name": "time_ms",
+                    "required": true,
+                    "type": "integer"
+                },
+                {
+                  "in": "body",
+                  "name": "hosCertifiedDailyLogParams",
+                  "required": true,
+                  "schema": {
+                    "properties": {
+                      "shippingDocIds": {
+                        "description": "One or more customer shipping document IDs to add to driver's daily certified log. Multiple values must be delimited by a comma or space. Max of 255 characters",
+                        "example": "101, 104",
+                        "type": "string"
+                      },
+                      "trailerNames": {
+                        "description": "One or more trailer names to add to driver's daily certified log. Multiple names must be delimited by a comma or space. Max of 255 characters",
+                        "example": "80111, 80112",
+                        "type": "string"
+                      }
+                    },
+                    "type": "object"
+                  }
+                }
+              ],
+              "responses": {
+                "200": {
+                  "description": "Returns the edited certified daily log",
+                  "schema": {
+                    "$ref": "#/definitions/HosCertifiedDailyLogResponse"
+                  }
+                },
+                "default": {
+                  "description": "Unexpected error.",
+                  "schema": {
+                    "$ref": "#/definitions/ErrorResponse"
+                  }
+                }
+              }
+            }
+        },
         "/fleet/hos_logs": {
             "post": {
                 "tags": [
@@ -4507,6 +4623,27 @@
               }
             }
           }
+        },
+        "HosCertifiedDailyLogResponse": {
+            "type": "object",
+            "properties": {
+                "dayStartMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "Unix epoch time (in ms) of time of when this certified daily log starts",
+                    "example": 1462881998034
+                },
+                "shippingDocIds": {
+                    "type": "string",
+                    "description": "One or more customer shipping document IDs to add to driver's daily certified log. Multiple values must be delimited by a comma or space. Max of 255 characters",
+                    "example": "101, 104"
+                },
+                "trailerNames": {
+                    "type": "string",
+                    "description": "One or more trailer names to add to driver's daily certified log. Multiple names must be delimited by a comma or space. Max of 255 characters",
+                    "example": "80111, 80112"
+                }
+            }
         },
         "HosLogsSummaryResponse": {
           "type": "object",


swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+                  "required": true,
+                  "schema": {
+                    "properties": {
+                      "timestamp": {
+                        "description": "A specified timestamp within the bounds of a driver's daily log, specified in milliseconds UNIX time.",
+                        "example": 1462881998034,
+                        "type": "integer"
+                      },
+                      "shippingId": {
+                        "description": "Shipping Id to add to driver's daily log.",
```
sskhandek: Should add restriction that it can only be a max of 255 characters

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+                        "example": 101,
+                        "format": "int64",
+                        "type": "integer"
+                      },
+                      "trailerName": {
+                        "description": "Trailer name to add to driver's daily log",
```
sskhandek: Should add restriction that it can only be a max of 255 characters

swagger.json
```diff
@@ -1121,6 +1121,77 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/hos_daily_log": {
+            "patch": {
+              "consumes": [
+                "application/json"
+              ],
+              "description": "Modify details, such as trailer number or shipping ID, for a driver's daily log",
```
prasad-shinde: probably a PM question, do we want to make it explicit here that we are going to append this new information rather than overwriting it?

prasad-shinde:
Verify it online

swagger.json
prasad-shinde: there is also a trick here that, If they are appending 255 chars to a field that already has 100 chars the last 100 chars from the request will be ignored.

We should talk with PM about this if we want to throw and error for scenarios like this or just truncate on our side which means the user is not aware that this happened on the backend

swagger.json
```diff
@@ -1121,6 +1121,77 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/hos_daily_log": {
+            "patch": {
```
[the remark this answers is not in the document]
silverspace: POST would be used for creating a new hos_daily_log, while PATCH would be used for updating an existing hos_daily_log.

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+              "consumes": [
+                "application/json"
+              ],
+              "description": "Modify details, such as trailer number or shipping ID, for a driver's daily log",
+              "operationId": "edit_fleet_drivers_hos_daily_log_details",
```
silverspace: Could we s/edit/patch?

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+              "parameters": [
+                {
+                  "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                  "description": "ID (integer) of the driver with HOS logs.",
+                  "in": "path",
+                  "name": "driver_id",
+                  "required": true,
+                  "type": "string"
```
silverspace: This should be type "integer".

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+                },
+                {
+                  "in": "body",
+                  "name": "editHosDailyLogParams",
+                  "required": true,
+                  "schema": {
+                    "properties": {
+                      "timestamp": {
```
silverspace: In most of our other APIs we use the key `timeMs`. Do we want to do the same here?

swagger.json
silverspace: Some more description here would be useful. It's not clear whether this is a samsara internal id or a customer specific id.

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+                        "description": "A specified timestamp within the bounds of a driver's daily log, specified in milliseconds UNIX time.",
+                        "example": 1462881998034,
+                        "type": "integer"
+                      },
+                      "shippingId": {
+                        "description": "Shipping Id to add to driver's daily log.",
+                        "example": 101,
+                        "format": "int64",
+                        "type": "integer"
```
silverspace: Do we require customers to use integer values for shippingId? Is this an identifier that they create? If so, it should be a string.

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+                  }
+                },
+                "default": {
+                  "description": "Unexpected error.",
+                  "schema": {
+                    "$ref": "#/definitions/ErrorResponse"
+                  }
+                }
+              },
+              "summary": "/fleet/drivers/{driver_id}/hos_daily_log",
```
silverspace: Could you add a regex for `driver_id` here?

swagger.json
```diff
@@ -1121,6 +1121,77 @@
...
+                        "type": "integer"
+                      },
+                      "trailerName": {
+                        "description": "Trailer name to add to driver's daily log",
+                        "example": "80111",
+                        "type": "string"
+                      }
+                    },
+                    "required": [
+                      "timestamp"
```
silverspace: Is this the same thing as `dayStartMs` in the response?

swagger.json
```diff
@@ -4508,6 +4579,27 @@
...
+                    "format": "int64",
+                    "description": "Unix epoch time (in ms) of time fo when this daily log starts",
+                    "example": 1462881998034
+                },
+                "shippingId": {
+                    "type": "string",
+                    "description": "Shipping Id, potentially multiple Ids delimited by a comma or space",
+                    "example": "1059984"
+                },
+                "trailerNumber": {
```
silverspace: The request body uses the key `trailerName`, but the response uses `trailerNumber`.

swagger.json
sskhandek: If I wanted to faithfully model how the UI does it, it would seem to make the most sense to have it overwritten since the data model is just one large string instead of a list. Perhaps this can be clearer in the description

swagger.json
```diff
@@ -1121,6 +1121,122 @@
...
+                "parameters": [
+                    {
+                      "$ref": "#/parameters/accessTokenParam"
+                    },
+                    {
+                      "description": "ID (integer) of the driver with HOS logs.",
+                      "in": "path",
+                      "name": "driver_id",
+                      "required": true,
+                      "type": "string"
```
dsharrison: Should this be an integer?

swagger.json
```diff
@@ -4508,6 +4624,27 @@
...
+            "properties": {
+                "dayStartMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "Unix epoch time (in ms) of time of when this daily log starts",
+                    "example": 1462881998034
+                },
+                "shippingDocNumbers": {
+                    "type": "string",
+                    "description": "Customer Shipping Document Numbers, potentially multiple numbers delimited by a comma or space",
```
dsharrison: nit: is `Customer Shipping Document Numbers` a proper noun? If not let's make it `Customer shipping document numbers`.

swagger.json
```diff
@@ -4508,6 +4624,27 @@
...
+                    "example": "1059984"
+                },
+                "trailerNames": {
+                    "type": "string",
+                    "description": "Trailer Number, potentially multiple names delimited by a comma or space",
```
dsharrison: nit: can we use `Trailer name` in the description rather than `Trailer Number`? This more closely aligns with the description of the patch parameter.

swagger.json
```diff
@@ -1121,6 +1121,122 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/hos_certified_day_chart/{time_ms}": {
```
silverspace: Do we want to expose this as a day chart? This is going to open up a can of worms for noon to noon fleets :(

sskhandek: I wrote it since I saw the term `day chart` being used in the legal lingo, but I'm happy to just keep it to `hos_certified_chart` if that sounds reasonable

swagger.json
```diff
@@ -1121,6 +1121,122 @@
...
+                    "required": true,
+                    "type": "integer"
+                },
+                {
+                  "in": "body",
+                  "name": "hosCertifiedDayChartParams",
+                  "required": true,
+                  "schema": {
+                    "properties": {
+                      "shippingDocNumbers": {
```
silverspace: Would `Id` make sense over `Number` here? We don't define what their shipping doc should look like, so it could be alphanumeric.

swagger.json
silverspace: Nit: could you make `chart` plural? (collections of resources should be plural form)

swagger.json
sskhandek: The official term is `Shipping Document Number` from the legal lingo which is why I was chose those words here. Do you still think Id is more appropriate?

swagger.json
sskhandek: Yes, I overlooked this. Thank you

swagger.json
silverspace: Do we expect this to always be a number? I think we should update our legal lingo since it's incorrect :)

swagger.json
silverspace: Is this really about certifying the driver's `shift`? Would `hos_certified_shift` make sense? 

In either case, I think we should document what value they need to pass for timeMs, and how we resolve that timestamp to their shift logs.

swagger.json
sskhandek: Sorry, I meant specifically from the ELD mandate: https://www.law.cornell.edu/cfr/text/49/appendix-A_to_subpart_B_of_part_395

swagger.json
sskhandek: Does the description I provided for `time_ms` parameter work to clarify that the timestamp should be in the bounds of a driver's shift? Or should I provide an extra description at the getter and patch level?

swagger.json
sskhandek: done

swagger.json
sskhandek: done

sskhandek:
@silverspace @dsharrison I made some more modifications to use the term `shift` instead of `day_chart` along with your other comments. Let me know what you all think!

swagger.json
```diff
@@ -1121,6 +1121,122 @@
...
+                    "default": {
+                        "description": "Unexpected error.",
+                        "schema": {
+                            "$ref": "#/definitions/ErrorResponse"
+                        }
+                    }
+                }
+            },
+            "patch": {
+              "summary": "/fleet/drivers/{driver_id:[0-9]+}/hos_certified_shifts/{time_ms:[0-9]+}",
```
faizandfurious: nit: There's a discrepancy between `get` request's `{time_ms}` vs the `patch` requests `{time_ms:[0-9]+}`. I think we should choose one and standardize.

swagger.json
```diff
@@ -4508,6 +4624,27 @@
...
+                "dayStartMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "Unix epoch time (in ms) of time of when this certified shift starts",
+                    "example": 1462881998034
+                },
+                "shippingDocIds": {
+                    "type": "string",
+                    "description": "Customer shipping document IDs, potentially multiple numbers delimited by a comma or space",
+                    "example": "1059984"
```
faizandfurious: nit: We may want the examples shown here match the examples within `hosCertifiedShiftParams`.

swagger.json
```diff
@@ -4508,6 +4624,27 @@
...
+                },
+                "trailerNames": {
+                    "type": "string",
+                    "description": "Trailer number, potentially multiple names delimited by a comma or space",
```
faizandfurious: These descriptions differ from the descriptions in `hosCertifiedShiftParams`, we should potentially standardize. If not, I think using "Trailer number" in this description is confusing, since the name of the field is "trailerNames".

swagger.json
```diff
@@ -1121,6 +1121,122 @@
...
+                        "name": "time_ms",
+                        "required": true,
+                        "type": "integer"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Returns a certified shift for the driver, if one exists",
+                        "schema": {
+                            "$ref": "#/definitions/HosCertifiedShiftResponse"
```
faizandfurious: Based on the name of the endpoint, I'd expect the response to include more than just `dayStartMs`/`shippingDocIds`/`trailerNames`.

swagger.json
sskhandek: You're right here, I keep shifting back and forth between names and I don't catch some of them

silverspace: This description also mentions both `number` and `multiple names`.

swagger.json
```diff
@@ -4508,6 +4624,27 @@
...
+            "properties": {
+                "dayStartMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "Unix epoch time (in ms) of time of when this certified daily log starts",
+                    "example": 1462881998034
+                },
+                "shippingDocIds": {
+                    "type": "string",
+                    "description": "Customer shipping document IDs, potentially multiple numbers delimited by a comma or space",
```
silverspace: Nit: s/numbers/identifiers

swagger.json
```diff
@@ -1121,6 +1121,122 @@
...
+                    "type": "integer"
+                },
+                {
+                  "in": "body",
+                  "name": "hosCertifiedShiftParams",
+                  "required": true,
+                  "schema": {
+                    "properties": {
+                      "shippingDocIds": {
+                        "description": "One or more customer shipping document IDs to add to driver's daily certified log. Multiple numbers can be inserted, delimited by a comma or space. Max of 255 characters",
```
silverspace: Do we really want to offer space separated value support? Can we just assert that multiple values have to be comma separated?

swagger.json
```diff
@@ -1121,6 +1121,122 @@
...
+              ],
+              "consumes": [
+                "application/json"
+              ],
+              "parameters": [
+                {
+                  "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                  "description": "ID (integer) of the driver with certified daily log.",
```
silverspace: This description is a bit confusing. Maybe `Identifier of the driver whose certified daily log is being updated`?

swagger.json
sskhandek: Resolved in person that we're going to scope this PR to just these fields

PR #119, opened 2019-01-24
Update driver info api endpoints (/fleet/drivers/...)

alberthu16:
preview link: https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/samsarahq/api-docs/1d4060fefcfdf0ae20def22621775f8206635080/swagger.json#operation/patchDriverById
jira: https://samsaradev.atlassian.net/browse/CM-423

- remove driver object in response for PUT /fleet/drivers/inactive/:id
- remove `isDeactivated` field from PATCH /fleet/drivers/:id (because we don't support modifying this field on PATCH)
- use `tagIds` in incoming requests (PATCH, POST), and return `tags` in responses (GET, PATCH, POST)
- match some driver fields to their gql variable names for consistency
- fix incorrect or missing examples

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -4280,10 +4280,6 @@
                 {
                     "type": "object",
                     "properties": {
-                        "isDeactivated": {
-                            "type": "boolean",
-                            "description": "True if the driver account has been deactivated."
-                        },
                         "currentVehicleId": {
                             "type": "integer",
                             "format": "int64",
--- a/swagger.json
+++ b/swagger.json
@@ -4124,7 +4124,7 @@
                 "username": {
                     "type": "string",
                     "description": "Driver's login username into the driver app.",
-                    "example": "Susan Jones"
+                    "example": "susanjones"
                 },
                 "phone": {
                     "type": "string",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -963,10 +963,7 @@
                 ],
                 "responses": {
                     "200": {
-                        "description": "Returns the reactivated driver, which is now available at /fleet/drivers/{driver_id}.",
-                        "schema": {
-                            "$ref": "#/definitions/CurrentDriver"
-                        }
+                        "description": "Successfully reactivated the driver, which is now referenced by /fleet/drivers/{driver_id | external_id}.",
```
AChelikani: Should the PUT return the updated object back?

swagger.json
```diff
@@ -4146,24 +4143,28 @@
...
                 "eldExemptReason": {
                     "type": "string",
-                    "description": "Reason that this driver is exempt from the Electronic Logging Mandate (see eldExempt)."
+                    "description": "Reason that this driver is exempt from the Electronic Logging Mandate (see eldExempt).",
+                    "example": "Some reason."
                 },
                 "usBigDayExemptionEnabled": {
                     "type": "boolean",
-                    "description": "Flag indicating this driver may use Big Day exemptions in HOS logs."
+                    "description": "Flag indicating this driver may use Big Day exemptions in HOS logs.",
```
AChelikani: Is it necessary to add something about "US" in the description? Just wondering since it's in the variable name.

swagger.json
```diff
@@ -4169,7 +4173,8 @@
                 },
                 "hosDayStartHour": {
                     "type": "integer",
-                    "description": "0 indicating midnight-to-midnight HOS driving hours, 12 to indicate noon-to-noon driving hours."
+                    "description": "0 indicating midnight-to-midnight HOS driving hours, 12 to indicate noon-to-noon driving hours.",
```
AChelikani: "indicating" -> "to indicate" for parallel wording?

swagger.json
alberthu16: The DELETE doesn't return updated object either, so I wanted to match that. Returning the updated object would also require some changes in the api.

swagger.json
alberthu16: Jonte gave me this description :)

PR #120, opened 2019-01-30
swagger: add preview docs for /hos/audit_logs

dsharrison:
Add preview docs for the undocumented `hos/audit_logs` API. These preview changes will be landed to help confirm the breaking backend change we want to make.

[Preview]


swagger.json
```diff
@@ -1519,6 +1573,52 @@
                 }
             }
         },
+        "/fleet/hos/audit_logs": {
```
silverspace: This should be at `/fleet/drivers/hos/audit_logs` for consistency with `/fleet/drivers/:id/hos/audit_logs`.

swagger.json
```diff
@@ -1519,6 +1573,52 @@
...
+                    {
+                        "name": "start_ms",
+                        "description": "Beginning of the time range, specified in milliseconds Unix epoch time.",
+                        "required": true,
+                        "in": "query",
+                        "type": "integer"
+                    },
+                    {
+                        "name": "end_ms",
+                        "description": "End of the time range, specified in milliseconds Unix epoch time.",
```
silverspace: This is probably a widespread issue across all of our endpoints, but we should specify whether `start_ms` and `end_ms` are inclusive or exclusive.

swagger.json
```diff
@@ -1519,6 +1573,52 @@
...
             }
         },
+        "/fleet/hos/audit_logs": {
+            "get": {
+                "tags": [
+                    "Fleet"
+                ],
+                "summary": "/fleet/hos/audit_logs",
+                "description": "Get the list of HOS (hours of service) audit logs for the specified time range.",
+                "operationId": "get_hos_audit_logs",
```
silverspace: Nit: s/get/list

swagger.json
```diff
@@ -5118,6 +5229,83 @@
...
+        "HosAuditLogBase": {
+            "type": "object",
+            "properties": {
+                "driverId": {
+                    "type": "integer",
+                    "description": "Driver ID",
+                    "example": 1234
+                },
+                "vehicleId": {
+                    "type": "integer",
```
silverspace: Can we add `format: int64` for these 2 integers?

swagger.json
```diff
@@ -5118,6 +5229,83 @@
...
+                    "enum": [
+                        "Created",
+                        "Deleted",
+                        "Updated",
+                        "Created (Auto Duty)",
+                        "Deleted (Auto Duty)",
+                        "Vehicle Assignment Created",
+                        "Vehicle Assignment Deleted",
+                        "Vehicle Assignment Updated",
+                        "Signed In",
```
silverspace: Thanks for making this an enum!!

swagger.json
```diff
@@ -1519,6 +1573,52 @@
                 }
             }
         },
+        "/fleet/hos/audit_logs": {
+            "get": {
+                "tags": [
+                    "Fleet"
+                ],
+                "summary": "/fleet/hos/audit_logs",
+                "description": "Get the list of HOS (hours of service) audit logs for the specified time range.",
```
silverspace: Nit: s/Get/List

swagger.json
```diff
@@ -5118,6 +5229,83 @@
...
+                "logStartMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "The time at which the log/HOS started in Unix epoch milliseconds.",
+                    "example": 1462881998034
+                },
+                "createdAtMs": {
+                    "type": "integer",
+                    "format": "int64",
+                    "description": "The time in Unix epoch milliseconds that the audit entry was created.",
```
silverspace: Should we add some additional context describing the difference between logStartMs and createdAtMs? Maybe we could describe that if the vehicle gateway loses cellular connectivity for 5 minutes, createdAtMs will be 5 minutes behind logStartMs?

swagger.json
```diff
@@ -1118,6 +1118,60 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/hos/audit_logs": {
```
silverspace: At a high level, should HOS logs and HOS audit logs be first class citizens in our system? I ask because they do have `ids` in our system, and belong to both drivers and vehicles. 

e.g. `/fleet/hos/audit_logs` with query parameter filters on `driver-ids` or `vehicle-ids`? This isn't something that we need to tackle in this PR, but it is something that we'll want to address in our Golden API v2 routes.

swagger.json
dsharrison: good call, I think this change is fine to make as we are making a change to the overall response as well.

swagger.json
dsharrison: Yes, the compliance team needs to spend some time working through what our first class citizens are. I agree that these records would make sense at the top level.

swagger.json
dsharrison: sure!

swagger.json
dsharrison: yep, this was a copy/paste from an out of date reference I think!

PR #121, opened 2019-01-30
swagger: Fix bug on POST fleet/drivers/:driver_id/documents

jenniferjzhang:
Previously, the documentation showed the `"fields"` parameter wrapped in 2 arrays:
screen shot 2019-01-30 at 1 04 15 pm

Now it should not be:
screen shot 2019-01-30 at 1 03 57 pm

This was brought up by Gordon when making documents.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -6185,17 +6185,6 @@
                 "$ref": "#/definitions/DocumentField"
             }
         },
-        "DocumentFieldCreates": {
-            "type": "array",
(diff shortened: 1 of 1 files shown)


ecshreve:
I'm still seeing the double array in the sample response, I think I see where the issue might be though, will leave another comment in a minute.

screen shot 2019-01-30 at 1 22 15 pm

PR #122, opened 2019-02-01
swagger: remove the incorrect AUTO_DUTY enum from audit_logs

dsharrison:
All logs returned from the audit logs API include the corresponding log status for statusType regardless of whether they were created via auto duty or a manual driver status change.


dsharrison:
@silverspace yes, we track the action source as auto-duty but the log status should simply be "DRIVING" or "OFF".

PR #123, opened 2019-02-01
swagger: Add preview for /fleet/drivers/{driver_id}/log_edits

prasad-shinde:
This endpoint will return the status of the log edits.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1118,6 +1118,59 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/log_edits": {
+            "get": {
+              "tags": [
+                "Fleet"
+              ],
+              "summary": "/fleet/drivers/{driver_id:[0-9]+}/log_edits",
+              "description": "Get a summary of the carrier edits for a specified driver. Note: This does not include edits made by driver",
+              "operationId": "getLogEdits",
+              "parameters": [
+                {
+                    "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                    "name": "driver_id",
+                    "in": "path",
+                    "required": true,
+                    "description": "ID (integer) of the driver with HOS logs.",
+                    "type": "number",
+                    "format": "int64"
+                },
+                {
+                "name": "created_at_start_ms",
+                    "in": "query",
+                    "required": true,
+                    "description": "Beginning of the time range (record creation), specified in milliseconds UNIX time.",
+                    "type": "number",
+                    "format": "int64"
+                },
+                {
+                    "name": "created_at_end_ms",
+                    "in": "query",
+                    "required": true,
+                    "description": "End of the time range (record creation), specified in milliseconds UNIX time.",
+                    "type": "number",
+                    "format": "int64"
+                }
+              ],
+              "responses": {
+                "200": {
+                  "description": "Log edits with status of the logs.",
+                  "schema": {
+                    "$ref": "#/definitions/LogEditResponse"
+                  }
+                },
+                "default": {
+                  "description": "Unexpected error.",
+                  "schema": {
+                    "$ref": "#/definitions/ErrorResponse"
+                  }
+                }
+              }
+            }
+        },
         "/fleet/drivers/{driver_id}/hos/audit_logs": {
             "get": {
                 "tags": [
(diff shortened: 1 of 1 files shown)


prasad-shinde:
Validated the json using the online validator here: https://editor.swagger.io/?_ga=2.224314525.782054557.1549056086-260734191.1548290935

swagger.json
```diff
@@ -1118,6 +1118,58 @@
...
+              "parameters": [
+                {
+                    "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                    "name": "driver_id",
+                    "in": "path",
+                    "required": true,
+                    "description": "ID (integer) of the driver with HOS logs.",
+                    "type": "string"
```
seward: Shouldn't this be integer?

prasad-shinde: you are right!

swagger.json
```diff
@@ -1118,6 +1118,58 @@
             }
           }
         },
+        "/fleet/drivers/{driver_id}/log_edits": {
```
sskhandek: Should `driver_id` be spinal case according to: https://paper.dropbox.com/doc/API-Best-Practices--AWwpYp50cquK5DmNE9lXnyfWAg-HNfOvJK77p8UIC5BUFO8S

swagger.json
```diff
@@ -1118,6 +1118,58 @@
...
+            "get": {
+              "tags": [
+                "Fleet"
+              ],
+              "summary": "/fleet/drivers/{driver_id:[0-9]+}/log_edits",
+              "description": "Get summary of log edits for a specified driver.",
```
dsharrison: We are only returning carrier edits here. Should we describe that either in the route or description?

swagger.json
prasad-shinde: nice! we now have some standards!

swagger.json
prasad-shinde: how about `Get a summary of carrier edits for a specified driver`

would you also prefer the endpoint name `/carrier_edit`?

prasad-shinde:
Updated the PR, ready for second pass

swagger.json
```diff
@@ -1118,6 +1118,58 @@
...
+              "parameters": [
+                {
+                    "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                    "name": "driver-id",
+                    "in": "path",
+                    "required": true,
+                    "description": "ID (integer) of the driver with HOS logs.",
+                    "type": "int64"
```
dsharrison: This should be type number, format int64.

prasad-shinde: fixed!

swagger.json
```diff
@@ -1118,6 +1118,59 @@
...
+              ],
+              "summary": "/fleet/drivers/{driver-id:[0-9]+}/log_edits",
+              "description": "Get a summary of the carrier edits for a specified driver. Note: This does not include edits made by driver",
+              "operationId": "getLogEdits",
+              "parameters": [
+                {
+                    "$ref": "#/parameters/accessTokenParam"
+                },
+                {
+                    "name": "driver-id",
```
seward: Could you make driver-id a parameter like accessTokenParam?  We use it so many places as a path param and a lot of the definitions are slightly inconsistent. ☹️   As an aside, it could probably be snake_case, since it's a path param and not a query param, but I think this is probably ok as spinal-case.

swagger.json
```diff
@@ -1118,6 +1118,59 @@
...
+                    "in": "path",
+                    "required": true,
+                    "description": "ID (integer) of the driver with HOS logs.",
+                    "type": "number",
+                    "format": "int64"
+                },
+                {
+                "name": "created_at_start_ms",
```
seward: created_at_start_ms and created_at_end_ms are query parameters and should be spinal-case.

prasad-shinde:
@seward I see we have some conventions defined where we want to move and I agree with that! I spoke with our team here and since we have some stuff already landed which uses the old convention, we are thinking of using the same conventions so that our customers don't have confusion. We are planning to move these api's to use the right conventions when we move to the `v2` version.

swagger.json
prasad-shinde: changed back to snake_case

swagger.json
```diff
@@ -4730,6 +4783,65 @@
             }
           }
         },
+        "LogEditResponse": {
+            "type": "object",
+            "properties": {
+                "log_edits": {
```
prasad-shinde: nts: change this to `logEdit`

PR #124, opened 2019-02-06
swagger certified_daily_logs: Param casing

sskhandek:
Changing the param casing of the `hos_certified_daily_logs` endpoint


sskhandek:
Closing because although `spinal-case` was listed as a best practice, it seems that we're going to keep snake case for the short term

PR #129, opened 2019-02-26
swagger.json: update /fleet/unassigned_driving_segments args

zkohl:
Update to swagger docs for /fleet/unassigned_driving_segments to add filterType. Currently only filtering on createdAtMs is supported, but in the future we can add support for querying on logMs.

Renames createdAtStartMs to startMs and createdAtEndMs to endMs so the args can make sense for filterying by logMs in the future.

Also removes Default tag.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1417,7 +1417,7 @@
         "/fleet/unassigned_driving_segments": {
             "get": {
                 "tags": [
-                    "Fleet", "Default"
+                    "Fleet"
                 ],
                 "summary": "/fleet/unassigned_driving_segments",
                 "description": "Get the unassigned driving segments for a specified range.",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -1427,20 +1427,27 @@
...
+                        "description": "End of the filter time range, specified in milliseconds UNIX time.",
                         "type": "integer",
                         "format": "int64"
+                    },
+                    {
+                        "name": "filterType",
+                        "in": "query",
+                        "required": true,
+                        "description": "Determines the property that the startMs/endMs time range filter applies to. Currently only 'createdAtMs' is supported",
+                        "type": "string"
```
dsharrison: Can we make this an enum to be clear that we only support a single value right now?

zkohl: sure, fixed

PR #130, opened 2019-02-27
add status to UnassignedDrivingSegment object

faizandfurious:


swagger.json
```diff
@@ -4122,6 +4122,11 @@
                     "format": "int64",
                     "example": 719
                 },
+                "status": {
+                  "type": "string",
+                  "description": "Status of assignment for this segment.",
+                  "example": "PENDING"
```
dsharrison: Can we make this an enum?

faizandfurious: 👍

PR #134, opened 2019-06-13
Add endMs as optional parameter to /v1/fleet/hos_logs_summary endpoint

faizandfurious:


katielee-samsara:
Following up here -  @faizandfurious would you be able to update the docs within the samsara.com/backend repo, and then post in the Slack channel for #api-facing-changes with the updates? 

Thanks!

PR #135, opened 2019-06-19
add beta docs folder

seward:
Adds a placeholder for the beta docs so we can add the beta docs to the swagger deploy pipeline.

The intention is that `swagger.json` will always be what our default public docs are based off, so when we release the beta API endpoints then we will...

* move the existing `swagger.json` to `v1/swagger.json`
* move `beta/swagger.json` to `swagger.json`
* remove the `beta` folder.

I'm creating this placeholder PR first so we can agree on the structure of the beta swagger before we add it to the deploy pipeline.

### DIFF
--- a/beta/README.md
+++ b/beta/README.md
@@ -1,0 +1,2 @@
+This is documentation for the unstable/future APIs release.
+Breaking changes may be introduced.


silverspace:
Are we planning on starting to update the beta swagger file now, or only after releasing the endpoints? I don't think we should start to publish the beta swagger file until we have critical mass, since folks will definitely notice a new file on master.

seward:
> Are we planning on starting to update the beta swagger file now, or only after releasing the endpoints? I don't think we should start to publish the beta swagger file until we have critical mass, since folks will definitely notice a new file on master.

Maybe a question for @katielee-samsara. I think it's fine to hold off on adding the beta APIs to the swagger deploy pipeline until we're prepared to release them officially. Then we can just cut over the old docs into a `v1/swagger.json` and replace `swagger.json` with the (formerly) beta swagger.

katielee-samsara:
I'm good to hold off on adding beta APIs until we're ready to release.

seward:
Closing and using a [separate branch](https://github.com/samsarahq/api-docs/tree/beta), as discussed.

PR #141, opened 2019-11-20
Jenn/uds status docs

jennkao:
Add dutyStatus field (int64) and segments field ([]UnassignedDrivingSegment) to UnassignedDrivingSegment. Also update "QUEUED" and "MULTIPLE" as possible statuses for UnassignedDrivingSegment.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -4254,9 +4254,11 @@
                         "PENDING",
                         "ACCEPTED",
                         "REJECTED",
+                        "QUEUED",
                         "UNASSIGNED",
                         "UNKNOWN",
-                        "ANNOTATED"
+                        "ANNOTATED",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -4254,15 +4254,29 @@
                         "PENDING",
                         "ACCEPTED",
                         "REJECTED",
+                        "QUEUED",
```
Calebs97: Do we need to add "ANNOTATED"?

jennkao: ah Faiz has an open PR for this: https://github.com/samsarahq/api-docs/pull/140

these changes shouldn't be merged until 12/10, so we should probably keep our changes apart in case we want to update docs before then

PR #143, opened 2021-04-12
workflows: adding the option to manually trigger github workflow

dgiridharan:
Adding the workflow_dispatch option to the Sync OAS to ReadMe workflow
so we can manually run the workflow when needed (for debugging)


.github/workflows/readme-api-sync.yml
```diff
@@ -1,5 +1,6 @@
 name: Sync OAS to ReadMe
 on:
+  workflow_dispatch:
```
seward, line 3: ooc how does this work?

seward (approved):

PR #144, opened 2021-04-13
workflows: change condition of readme sync to happen on doc changes

dgiridharan:
Add extra conditions to the readme sync workflow to happen only when
swagger-readme.json or any files under markdown directory is changed
on the master branch.

### DIFF
--- a/.github/workflows/readme-api-sync.yml
+++ b/.github/workflows/readme-api-sync.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+    paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/readme-api-sync.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+    paths:
```
AChelikani, line 7: Nice

AChelikani (approved):

PR #148, opened 2021-04-22
OpenAPI specification fixes for AddressType and int64

softdevca:
There are two fixes required to the OpenAPI specification. F

First is that the container of enums shouldn't be array because it causes the type to be interpreted as `Seq[MyEnum]`.  This causes OpenAPI Generator to create code that won't compile.

Second is that `int64` should be consistently have the `integer` type rather than the more general `number` type. The `number` type causes the OpenAPI generator to use the `BigDecimal` type for what should be a primitive and fail to compile

Tested with the scala-* generators.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -9649,7 +9649,6 @@
       "properties": {
         "addressTypes": {
           "description": "Reporting location type associated with the address (used for ELD reporting purposes).",
-          "type": "array",
           "items": {
             "type": "string",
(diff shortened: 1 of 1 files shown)


softdevca:
I understand the OpenAPI specification for Samsara is not officially supported for code generation, but it would be a nice bonus if minor errors affecting code generation could be fixed.

PR #149, opened 2021-05-24
scripts: adds readme_changelog script to format and publish changelog posts to readme.

seward:
Adds a script to read local changelog files and  format & publish changelog posts to readme.

Also sets up the initial versions of those files:
| file | description |
| -- | -- |
| changelog.json | the combined changelog of all API versions |
| changelog_index.json | the index of changelog entry ID to readme io slug | 'ignored' to indicate which changelogs do not need to be posted |
| operation_ids.json | a generated mapping of <HTTP method, route template> to swagger operation ID so we can format the changelog post with correct references |


scripts/readme_changelog.py
```diff
@@ -0,0 +1,162 @@
...
+    return index
+
+"""
+Determines which changelogs should be published in this run.
+"""
+def diffChangelogs(index, changelogs):
+    diff = []
+    for k,c in changelogs.items():
+        if k in index:
+            print("already logged or ignored this changelog: {}".format(k))
```
seward, line 124: remove spurious print

.github/workflows/readme-changelog-sync.yml
```diff
@@ -0,0 +1,57 @@
+name: Sync Changelogs to ReadMe
+on:
+  workflow_dispatch:
```
dgiridharan, line 3: Thanks for adding this so we can kick this off on demand!

dgiridharan (approved):

.github/workflows/readme-changelog-sync.yml
```diff
@@ -0,0 +1,57 @@
...
+  push:
+    branches:
+      - master
+    paths:
+      - 'changelog.json'
+      - 'changelog_index.json'
```
seward, line 9: can probably drop `changelog_index` and `operation_ids` from here and just be `changelog.json`

PR #152, opened 2021-05-27
scripts changelog: add logging for workflow action

seward:
Adds logging for the changelog workflow action and corrects the environment variable name for the token.


vyhuynh8 (approved):
This looks good to me. All of the changes make sense and would help with debugging in the future.

